PR #1, opened 2017-02-03
Feature/model

malaffoon:
This is the initial checkin of the common model and messaging modules. I wanted two modules so i could proof out the multi-module handling of gradle. Yes, it uses gradle with nary a pom in sight. I'm new to gradle, please inspect carefully. 

I don't have full test coverage as i was more focused on getting the project framework in place. I did enable a coverage report as part of the build, so i can work on it in subsequent checkins.

If somebody wants to pull down this branch and see if they have any issues opening it in IntelliJ, i'd appreciate it. Might be an easier way to review the code anyway.


agorina:
Looks good to me.

PR #3, opened 2017-02-14
add tenancy chain util

malaffoon:


utils/src/main/java/rdw/utils/TenancyChain.java
```diff
@@ -0,0 +1,181 @@
...
+     * exam service.
+     *
+     * @param roleName role name, e.g. "ASMTDATALOAD"; case is ignored
+     * @return true if role is granted at any level
+     */
+    public boolean hasRole(final String roleName) {
+        return grants.stream().map(Grant::getRoleName).anyMatch(roleName::equalsIgnoreCase);
+    }
+
+    /**
```
malaffoon: Not sure what methods will be needed so i just put in what i need for now. I have a feeling this will be reworked a bit to make the different grant levels more loosely bound.

utils/src/test/java/rdw/utils/TenancyChainTest.java
```diff
@@ -0,0 +1,65 @@
...
+        final TenancyChain chain = TenancyChain.fromString(value);
+        assertThat(chain.isEmpty()).isTrue();
+        assertThat(chain.toString()).isEqualTo(value);
+    }
+
+    @Test
+    public void itShouldSupportASimpleClientGrant() {
+        final String value = "||ASMTDATALOAD|||SBAC|||||||||||||";
+        final TenancyChain chain = TenancyChain.fromString(value);
+        assertThat(chain.hasRole("ASMTDATALOAD")).isTrue();
```
malaffoon: This is pretty much what i needed.

PR #4, opened 2017-02-15
add user-related headers

malaffoon:


messaging/src/main/java/rdw/messaging/RdwMessageHeaderAccessor.java
```diff
@@ -6,9 +6,6 @@
 /**
  * A {@link org.springframework.messaging.support.MessageHeaderAccessor}
  * implementation giving access to RDW-specific headers.<br/>
- *
- * In theory, this could be refactored so each step of the ingest pipeline has its own set of
- * headers and accessor. However, this approach is sufficient for now.
  */
 public class RdwMessageHeaderAccessor extends MessageHeaderAccessor {
```
malaffoon: added setters/getters for new headers 
fleshed out javadocs

messaging/src/main/java/rdw/messaging/RdwMessageHeaders.java
```diff
@@ -13,9 +13,12 @@
  */
 public class RdwMessageHeaders {
```
malaffoon: cleaned up scope, none of this needs to be visible outside the package, that's the whole point of the accessor

malaffoon:
It seems Fairway doesn't have a lot of tools in their CI tool chain. In all my past lives i've used static code analysis tools both at the developer level and in CI. In one job, we even rejected builds if there were too many warnings from those tools. I'd love to have a couple of them available and hooked up to our builds.

That said, i will fight the whole PREFIX vs. Prefix thing. ALL_UPPERCASE_CONSTS is a throwback to the 1990's and languages with anemic tools. It is hard to type, harsh to read, and no longer adds any value since the tools highlight consts for you. SimpleCamelCasing is easier to type and read.

PR #6, opened 2017-02-23
DWR-139 - adding configuration bean for YAML parsing of yaml config f…

rcordeau:
…iles.

---
This change is [Reviewable](https://reviewable.io/reviews/smarterapp/rdw_common/6)

### DIFF
--- a/utils/build.gradle
+++ b/utils/build.gradle
@@ -1,0 +1,6 @@
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+dependencies {
+    compile ('org.springframework:spring-context:+')
+}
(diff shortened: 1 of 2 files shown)


utils/build.gradle
```diff
@@ -0,0 +1,6 @@
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
```
malaffoon: this is in the parent gradle file already

utils/build.gradle
```diff
@@ -0,0 +1,6 @@
...
+
+dependencies {
+    compile ('org.springframework:spring-context:+')
```
malaffoon: this is fine for now, but i will be tempted to have something like rdw-spring-utils at some point

utils/build.gradle
rcordeau: Yeah - I realized that after pushed it up that it was redundant.

utils/build.gradle
rcordeau: I needed it because of the @Configuration annotation

PR #8, opened 2017-02-24
add unit test for yaml configurer

malaffoon:


utils/src/test/resources/test.sql.yml
```diff
@@ -0,0 +1,10 @@
+widget:
+  getWidgets:
+    sql: >-
+      select * from widget
+  getWidget:
+    sql: >-
+      select * from widget where id=:id
+  getClip:
```
malaffoon: this is just demonstrating that, without `>-` the trailing newlines will be retained in the sql

PR #9, opened 2017-03-01
add special setters to support easy CSV import

malaffoon:


agorina:
I am not sure I see how we could get the data in, we need the item level data. Let's chat tomorrow.

PR #11, opened 2017-03-07
add hasRoleForXxx methods

malaffoon:
I think this is what you need; let me know.


utils/src/test/java/org/opentestsystem/rdw/utils/TenancyChainTest.java
```diff
@@ -16,30 +16,47 @@ public void itShouldSupportAnEmptyChain() {
...
     public void itShouldSupportASimpleClientGrant() {
-        final String value = "||ASMTDATALOAD|||SBAC|||||||||||||";
+        final String value = "||ASMTDATALOAD|CLIENT||SBAC|||||||||||||";
         final TenancyChain chain = TenancyChain.fromString(value);
-        assertThat(chain.hasRole("ASMTDATALOAD")).isTrue();
         assertThat(chain.toString()).isEqualTo(value);
+        assertThat(chain.hasRole("ASMTDATALOAD")).isTrue();
+        assertThat(chain.hasRoleForClient("ASMTDATALOAD", "SBAC")).isTrue();
+        assertThat(chain.hasRoleForClient("SAREXTRACTS", "SBAC")).isFalse();
+        assertThat(chain.hasRoleForClient("ASMTDATALOAD", "FUBAR")).isFalse();
```
agorina: What is Client? Would I need to use it?

utils/src/main/java/org/opentestsystem/rdw/utils/TenancyChain.java
```diff
@@ -91,17 +91,73 @@ public boolean hasRole(final String roleName) {
...
+     *
+     * @param role role id or name
+     * @param school school id or name
+     * @return true if role is granted for given school
+     */
+    public boolean hasRoleForSchool(final String role, final String school) {
+        return grants.stream().anyMatch(grant -> "INSTITUTION".equalsIgnoreCase(grant.getLevel())
+                && (role.equalsIgnoreCase(grant.getRoleId()) || role.equalsIgnoreCase(grant.getRoleName()))
+                && (school.equalsIgnoreCase(grant.getSchoolId()) || school.equalsIgnoreCase(grant.getSchoolName())));
     }
```
agorina: Is this the one I would need to call wit the role "ASMTDATALOAD" ?

Is school unique across States?

utils/src/test/java/org/opentestsystem/rdw/utils/TenancyChainTest.java
malaffoon: I just included it because the tenancy chain supports it.

utils/src/main/java/org/opentestsystem/rdw/utils/TenancyChain.java
malaffoon: It is currently coded assuming that school ids are unique across districts; we can always tweak it. You would probably use the hasRole(role, state, district, school) method.

PR #13, opened 2017-03-14
getbest

malaffoon:
Alla, i profiled this a bit. It is pretty darn fast ... for 6 attributes it was faster making the six getBestAttribute calls (60us) than looping through all the attributes, doing a switch statement, and favoring FINAL over INITIAL (75us). That was with 50 total attributes. Perhaps if i continued up to extracting all 20-25 attributes it might be faster doing the loop+switch but it definitely isn't a slam-dunk ... and this makes the client code a lot nicer to read.

Full disclosure: my first couple iterations of the getBestXxx methods were slower. A stream() version of it was taking 50-60 ms (i.e. 1000x slower)!


model/src/test/resources/TDSReport.sample2.xml
```diff
@@ -6,7 +6,7 @@
         <ExamineeAttribute context = "INITIAL" name = "FirstName" value = "LoadTest" contextDate = "2015-01-09T22:13:30.631" />
         <ExamineeAttribute context = "INITIAL" name = "Sex" value = "M" contextDate = "2015-01-09T22:13:30.631" />
         <ExamineeAttribute context = "INITIAL" name = "GradeLevelWhenAssessed" value = "11" contextDate = "2015-01-09T22:13:30.631" />
-        <ExamineeAttribute context = "INITIAL" name = "LastOrSurname" value = "Student0000072" contextDate = "2015-01-09T22:13:30.631" />
```
malaffoon: just some changes to make sure it was favoring FINAL over INITIAL

model/src/main/java/org/opentestsystem/rdw/model/Examinee.java
```diff
@@ -446,6 +446,48 @@ public void setContextDate(Calendar value) {
     // ************************************************************************
     // ************************************************************************
     // ************************************************************************
+
+    // Helpers for accessing attributes and relationships
+
+    /**
+     * Return attribute that matches given name, preferring FINAL over INITIAL.
+     * Algorithm may look clunky, but it's fast; don't change unless you prove it is faster.
+     * Yeah, and returning null is a lot faster than returning Optional, get over it.
```
agorina: Thank you for testing it and putting this comment :)

model/src/test/java/org/opentestsystem/rdw/model/TDSReportTest.java
```diff
@@ -31,4 +31,26 @@ public void thisTestEnforcesModelClasses() {
         final List<ToolUsage> toolUsages = tdsReport.getToolUsage();
         assertThat(toolUsages).isNotNull();
     }
+
+    @org.junit.Test
+    public void itShouldGetBestAttribute() {
+        final TDSReport tdsReport = XmlUtils.tdsReportFromXml(this.getClass().getResourceAsStream("/TDSReport.sample2.xml"));
```
agorina: Would we want to time this test to make sure it is fast. I mean for the future reference if somebody decided to change something?

malaffoon: I had a big ol' performance test that tested a bunch of stuff ... i chose not to check it in. Future devs will just have to set up their own performance test!

agorina: I was referring to something like this: 
```
@Timed(millis=1000)
```

PR #15, opened 2017-03-18
Change to java 8 time

malaffoon:
The default code generation had mapped some of the xsd:dateTime fields to Calendar, some to String. This makes them all Instant. Except effectiveDate which is a simple Date.


model/build.gradle
```diff
@@ -1,3 +1,3 @@
 dependencies {       
-  compile "com.google.code.gson:gson:2.8.0" 
+  compile 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.8.7'
```
rcordeau: You could remove...

PR #17, opened 2017-04-12
binary http resource resolver

malaffoon:
Moving this helper into common so it can be used in multiple projects.


model/src/main/resources/TestResultsTransmissionFormat.xsd
```diff
@@ -85,17 +85,14 @@ governing permissions and limitations under the License.
                   <xs:attribute name="formId"/>
```
malaffoon: Ok, these unrelated changes shouldn't be included.

PR #18, opened 2017-04-14
Update to TRT Schema version 1.10

malaffoon:


model/src/main/java/org/opentestsystem/rdw/model/Opportunity.java
```diff
@@ -245,6 +246,8 @@
     @XmlAttribute(name = "status", required = true)
     @XmlJavaTypeAdapter(CollapsedStringAdapter.class)
     private String status;
+    @XmlTransient
```
malaffoon: This seems to be the best way to handle optional fields that have a default value; instead of directly manipulating the field, force jaxml to go through the set/get methods.

model/src/main/java/org/opentestsystem/rdw/model/Opportunity.java
```diff
@@ -618,6 +621,31 @@ public void setStatus(String value) {
...
+     * Sets the value of the validity property.
+     *
+     * @param value
+     *     allowed object is
+     *     {@link String }
+     *
+     */
+    @XmlAttribute(name = "validity")
+    public void setValidity(String value) {
+        this.validity = value == null || value.isEmpty() ? "valid" : value;
```
jtreuting: `"valid"` is used here and up above to set the default, might be nice to have a const for it

model/src/main/java/org/opentestsystem/rdw/model/Opportunity.java
```diff
@@ -2024,8 +2051,9 @@ public String getType() {
              *     {@link String }
              *
              */
+            @XmlAttribute(name = "type")
             public void setType(String value) {
-                this.type = value;
+                this.type = value == null || value.isEmpty() ? "value" : value;
```
jtreuting: same note here

model/src/main/java/org/opentestsystem/rdw/model/Opportunity.java
malaffoon: Will do.

PR #20, opened 2017-04-18
Feature/security

esotrope:
Adds common security module with common permission service abstraction

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@ subprojects {
     compile 'org.slf4j:slf4j-api:1.7.22'
 
     testCompile 'junit:junit:4+'
+    testCompile 'org.mockito:mockito-core:2+'
     testCompile 'org.assertj:assertj-core:3.6.2'
   }
 
--- a/security/build.gradle
+++ b/security/build.gradle
@@ -1,0 +1,3 @@
+dependencies {
+    compile project(':rdw-common-webclient')
+}
--- a/security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
+++ b/security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
@@ -43,8 +43,10 @@ public class ComponentPermissionService implements PermissionService {
 
 		final Response<Role> response = client.getRole(component);
 
-		if (!response.getStatus().equals(Response.Status.SUCCESS)) {
-			throw new PermissionServiceException(String.format("Error getting roles and permissions for component \"%s\"", component));
+		if (response.getStatus() == null || !response.getStatus().equals(Response.Status.SUCCESS)) {
+			throw new PermissionServiceException(String.format(
+				"Error getting roles and permissions for component \"%s\". Server responded with message: \"%s\"",
+				component, response.getMessage()));
 		}
 
 		final Map<String, Collection<String>> permissionsByRole = newHashMap();
--- a/security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
+++ b/security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
@@ -1,0 +1,63 @@
+package org.opentestsystem.rdw.security.service;
+
+
+import org.opentestsystem.rdw.webclient.permissionservice.Component;
+import org.opentestsystem.rdw.webclient.permissionservice.Permission;
+import org.opentestsystem.rdw.webclient.permissionservice.PermissionWebServiceClient;
+import org.opentestsystem.rdw.webclient.permissionservice.Response;
+import org.opentestsystem.rdw.webclient.permissionservice.Role;
+
+import javax.validation.constraints.NotNull;
+import java.util.Collection;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Maps.newHashMap;
+import static com.google.common.collect.Sets.newHashSet;
+
+/**
+ * This permission service implementation integrates with the https://github.com/SmarterApp/SS_Permissions
+ * project and accesses information for one registered component.
+ */
+public class ComponentPermissionService implements PermissionService {
+
+	/**
+	 * Used to integrate with the permission service
+	 */
+	private PermissionWebServiceClient client;
+
+	/**
+	 * The component for which permission information will be looked up
+	 */
+	private String component;
+
+	public ComponentPermissionService(@NotNull final PermissionWebServiceClient client, @NotNull final String component) {
+		this.client = checkNotNull(client);
+		this.component = checkNotNull(component);
+	}
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public Map<String, Collection<String>> getPermissionsByRole() throws PermissionServiceException {
+
+		final Response<Role> response = client.getRole(component);
+
+		if (!response.getStatus().equals(Response.Status.SUCCESS)) {
+			throw new PermissionServiceException(String.format("Error getting roles and permissions for component \"%s\"", component));
+		}
+
+		final Map<String, Collection<String>> permissionsByRole = newHashMap();
+		for (Role role : response.getValue()) {
+			final Collection<String> permissions = newHashSet();
+			for (Component mapping : role.getMappings()) {
+				for (Permission permission : mapping.getPermissions()) {
+					permissions.add(permission.getName());
+				}
+			}
+			permissionsByRole.put(role.getRole(), permissions);
+		}
+		return permissionsByRole;
+	}
+
+}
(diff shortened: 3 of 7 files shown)


security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
```diff
@@ -0,0 +1,65 @@
...
+
+	/**
+	 * The component for which permission information will be looked up
+	 */
+	private String component;
+
+	public ComponentPermissionService(@NotNull final PermissionWebServiceClient client, @NotNull final String component) {
+		this.client = checkNotNull(client);
+		this.component = checkNotNull(component);
+	}
```
agorina: Do you really need both NotNull and checkNotNull?

jtreuting: I'm pretty sure @NotNull just described intent but doesn't actually do anything.  More to document.

security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
```diff
@@ -0,0 +1,65 @@
...
+
+	/**
+	 * {@inheritDoc}
+	 */
+	public Map<String, Collection<String>> getPermissionsByRole() throws PermissionServiceException {
```
agorina: Is there a reason this method returns permissions' names? If yes, then I would suggest renaming the method to *getPermissionsNamesByRole* it to match the functionality.

security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
```diff
@@ -0,0 +1,65 @@
...
+
+		final Response<Role> response = client.getRole(component);
```
agorina: Can response be null?

security/src/test/java/org/opentestsystem/rdw/security/service/ComponentPermissionServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+	private ComponentPermissionService service;
+
+	@Before
+	public void before() {
+		mockClient = mock(PermissionWebServiceClient.class);
+		service = new ComponentPermissionService(mockClient, component);
+	}
+
+	@Test(expected = NullPointerException.class)
+	public void constructor_should_throw_exception_for_null_parameters() throws Exception {
```
agorina: Nit: I suggest using a camel case for the test name to be consistent with the rest of the code.

security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
agorina: Thank you! What I am trying to understand if we like using ```@NotNull ``` annotations? I see some people doing it and some do not. Just curious.

Also, for _checkNotNull_, I would suggest adding a message what is checked: ```checkNotNull(client, "client may not be null")```

security/build.gradle
```diff
@@ -0,0 +1,3 @@
+dependencies {
+    compile project(':rdw-common-webclient')
```
malaffoon: After thinking about this last night ... 

Is there a reason to separate these? Will the `webclient` be used for anything other than supporting the security module? I think i'd just lump them together until such time as we find a need to separate them. That would also help with the naming issue (i don't think `webclient` is descriptive enough since it is only for permissions).

build.gradle
```diff
@@ -17,6 +17,7 @@ subprojects {
     compile 'org.slf4j:slf4j-api:1.7.22'
 
     testCompile 'junit:junit:4+'
+    testCompile 'org.mockito:mockito-core:2+'
```
malaffoon: have to keep an eye on this ... spring test brings in mockito-core:1.x

security/src/test/java/org/opentestsystem/rdw/security/service/ComponentPermissionServiceTest.java
malaffoon: i'll weigh in that this is not a nit, please rename.

security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.validation.constraints.NotNull;
+import java.util.Collection;
+import java.util.Map;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Maps.newHashMap;
+import static com.google.common.collect.Sets.newHashSet;
+
+/**
+ * This permission service implementation integrates with the https://github.com/SmarterApp/SS_Permissions
```
malaffoon: Is "integrates" the right word? Perhaps "This permission service implementation uses the permissions model described in the https://github.com/SmarterApp/SS_Permissions project ..."

security/src/test/java/org/opentestsystem/rdw/security/service/ComponentPermissionServiceTest.java
jtreuting: I think we should have a standard practice.  Have we been doing something like `shouldThrowExceptionWhenNullConstructorParameters`, meaning always starting with NULL.  This might be a flame war not sure.  I don't have a strong opinion other than trying to be consistent.

security/build.gradle
esotrope: Yeah it's thinking ahead. The webclient bucket is meant to hold all external REST/SOAP integration pieces. The security.service layer wraps this and provides the methods we think we will need from those APIs but if someone wants something else they could go directly to the webclient layer and talk to the APIs directly was my thinking

build.gradle
esotrope: Does it? I tried to static import mock() and it found nothing hence the inclusion.
I'll play with that today.

security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
esotrope: maybe communicates, talks, interfaces, consumes? any suggestion?

security/src/test/java/org/opentestsystem/rdw/security/service/ComponentPermissionServiceTest.java
esotrope: Ok I'll rename. Have we ever used an automated stylecheck as part of the build process? If we could automate it we can standardize this stuff more easily I'm thinking. Then the checker will nitpick for us : j

security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
esotrope: The [permission API controller](https://github.com/SmarterApp/SS_Permissions/blob/master/Permissions/src/main/java/org/opentestsystem/shared/permissions/rest/PresentationHandler.java) doesn't ever return null but I think the service invocation can throw runtime exceptions

security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
esotrope: The thought behind this that a permission is just a name or string literal. The remote permission service api wraps the name in a Permission object but there are no other fields describing a permission at the moment it seems, and our permission service consumers aren't interested in anything but the string literal permission id/name.

security/src/main/java/org/opentestsystem/rdw/security/service/ComponentPermissionService.java
esotrope: The idea here is to ensure that the object cannot be constructed in a non-functioning state. The @NotNull is to inform the object-creator that they cannot pass null - I think this also allows the IDE to communicate to the developer that a null argument pass will cause a null pointer exception.
What are your guys thoughts on these points?

malaffoon: For a class in a utility jar, i think this is a good approach. The @NotNull let's the IDE display warnings while the checkNotNull fast-fails. If this were an autowired service inside a spring boot app i would suggest neither is needed but this class can't know how it is being instantiated.

agorina: I think that for the autowired components/services this check is redundant. 

For all other instances:
- I see @NotNull as a matter of a personal style and also would like to know what other people think about it.
- _checkNotNull_ is the right thing to have in this case.

security/build.gradle
malaffoon: It's not a hill i'm going to die on but I'm going to stick to my suggestion: put them together for now and separate later if the need ever arises.

PR #22, opened 2017-04-19
spring version stuff

malaffoon:
This seems to fix the build problem.


build.gradle
```diff
@@ -1,7 +1,33 @@
+buildscript {
```
malaffoon: Doing the whole spring-dependencies thing here.

build.gradle
```diff
@@ -10,15 +36,22 @@ subprojects {
...
 
-  group = "org.opentestsystem.rdw.common"
-  version = "0.0.1-SNAPSHOT"
+  dependencyManagement {
+    dependencies {
+      dependency 'com.google.guava:guava:21.0'
+    }
+    imports {
+      // make springframework versions consistent with other projects in the RDW suite
+      mavenBom("org.springframework.boot:spring-boot-dependencies:1.5.2.RELEASE")
```
malaffoon: Use this to get versions; remove *all* explicit versions elsewhere.

PR #24, opened 2017-04-20
TRT changes based on ETS testing

malaffoon:
Also tweaked TenancyChain to match Landon's recent clarifications.


model/src/main/java/org/opentestsystem/rdw/model/Examinee.java
```diff
@@ -125,10 +91,10 @@ public void setKey(Long value) {
      *
      * @return
      *     possible object is
-     *     {@link Short }
+     *     {@link Boolean }
      *
      */
-    public Short getIsDemo() {
+    public Boolean getIsDemo() {
```
jtreuting: I'm assuming that means this is optional since it could be null?  Does it make sense to return `false` when `isDemo` is null?  that way the consumer doesn't need to figure out what a `null` value represents?  I don't know how this is used so not sure.

malaffoon: This is supposed to be a dumb data model class that just represents what was marshaled. So null is valid (i.e. isDemo attribute was absent). The savvy client will wrap this in business logic and do the null check.

If the `isDemo` attribute had a default value in the schema, i would treat this differently.

jtreuting: works for me

PR #25, opened 2017-04-20
fixed grant validation and tests

esotrope:


security/src/main/java/org/opentestsystem/rdw/security/support/Grants.java
```diff
@@ -23,10 +25,11 @@ private Grants() {}
 		if (grant == null) {
 			throw new NullPointerException("grant must not be null");
```
jtreuting: I know this isn't part of this PR.  Seems like we might not need this check, since letting it go will result in an NPE anyways ... not a big deal, just noticed it.

security/src/main/java/org/opentestsystem/rdw/security/support/Grants.java
```diff
@@ -23,10 +25,11 @@ private Grants() {}
...
 		}
-		final String[] tokens = grant.split("\\|");
-		if (tokens.length != 18) {
-			throw new IllegalArgumentException(
-				String.format("Error tokenizing tenancy chain grant. Grant \"%s\" does not container 17 \"|\" delimiters", grant));
+		final String[] tokens = grant.split("\\|", -1);
+		if (tokens.length != REQUIRED_NUMBER_OF_TENANCY_CHAIN_GRANT_DELIMITERS + 1) {
+			throw new IllegalArgumentException(String.format(
+				"Error tokenizing tenancy chain grant. Grant \"%s\" does not container %s \"|\" delimiters",
```
jtreuting: should say `does not contain` instead of `container`

esotrope: good catch -thanks

PR #26, opened 2017-04-26
add equals/hashCode for (old) TenancyChain.

malaffoon:
The only reason i need this is to simplify some tests in RDW_Ingest.


utils/src/test/java/org/opentestsystem/rdw/utils/TenancyChainTest.java
```diff
@@ -60,6 +63,34 @@ public void itShouldSupportMultipeRolesMultipleLevels() {
...
+        assertThat(chains).hasSize(2);
+        chains.add(TenancyChain.fromString("|SBAC|ASMTDATALOAD|CLIENT|SBAC||||||||||||||"));
+        assertThat(chains).hasSize(2);
+    }
+
+    @Test
+    public void itHashesBasedOnString() {
+        final String value = "|SBAC|ASMTDATALOAD|CLIENT|SBAC||||||||||||||";
+        final TenancyChain chain = TenancyChain.fromString(value);
+        assertThat(chain.hashCode()).isEqualTo(value.hashCode());
```
malaffoon: what do you think of this behavior?

jtreuting: that seems odd to me

malaffoon: I thought about it some more and refreshed my memory on the contract for equals/hashCode. It is perfectly fine for two objects (even of different classes) to have the same hashCode, equals will take care of things.

esotrope: sort of on topic, but not exactly - just wondering why this model parses the grant string and then puts it back together in the toString() method. Why not just keep a copy of the original encoded grant string and return that from toString() and reference a field in hashCode/equals instead of the toString() method. I have never seen toString() be invoked in hashCode/equals before - it seems really odd to me. I can imagine someone wanting to change to string to make it prettier looking and it messing with the hashcode/equals logic

malaffoon: I could keep the string intact and just search whenever i needed a value ... seems like it would complicate the code and would certainly be less performant if a tenancy chain is queried a bunch during its lifetime. Also, when i originally wrote this i thought we might be building tenancy chains programmatically which would require separate fields. I also started to add enum(s) for a couple fields which would require parsing (kind of like your class).

And, yeah, if somebody changes toString, they would have to modify equals/hashCode. I suppose i could pull out a helper to create the string and then invoke it from all the methods (i'll update the PR in a few).

utils/src/main/java/org/opentestsystem/rdw/utils/TenancyChain.java
```diff
@@ -159,6 +161,18 @@ public boolean hasRoleForSchool(final String role, final String school) {
                 && (school.equalsIgnoreCase(grant.getSchoolId()) || school.equalsIgnoreCase(grant.getSchoolName())));
     }
 
+    @Override
+    public boolean equals(final Object that) {
+        if (this == that) return true;
+        if (that == null || getClass() != that.getClass()) return false;
+        return Objects.equals(grants, ((TenancyChain)that).grants);
```
malaffoon: Changed it to be more conventional. Also saves constructing a String just to do equals/hashCode.

PR #27, opened 2017-05-03
move ArchiveService to RDW_Common

malaffoon:
This is in anticipation of its use for storing pdf batch files.

Mostly this is just moving files from one project to another. 
I did change the signature so the content is `byte[]` instead of `String`. Probably that will change again (or be added to) to support Streams or Files.


archive/src/test/java/org/opentestsystem/rdw/archive/S3ArchiveServiceTest.java
```diff
@@ -0,0 +1,59 @@
...
+@TestPropertySource(properties = {
+        "archive.root=s3://rdw-devtest/",
+        "cloud.aws.region.auto: false",
+        "cloud.aws.region.static: us-west-2"
+})
+@Import(value = {
+        ContextResourceLoaderConfiguration.class,  // needed to get AmazonS3 client
+        ArchiveConfiguration.class
+})
+public class S3ArchiveServiceTest {
```
malaffoon: Hmm, i wonder if this should be S3ArchiveServiceIT since it calls out to S3. It was really fast so i didn't notice.

rcordeau: I would agree - IT

archive/src/test/java/org/opentestsystem/rdw/archive/S3ArchiveServiceTest.java
```diff
@@ -0,0 +1,59 @@
...
+    private S3ArchiveService service;
+
+    @Test(expected = IllegalArgumentException.class)
+    public void itRequiresAS3Scheme() {
+        new S3ArchiveService("/tmp");
+    }
+
+    @Test
+    public void itShouldWriteAndReadContent() throws UnsupportedEncodingException {
+        final byte[] content = "TestyMcTestFace".getBytes("UTF-8");
```
rcordeau: Funny

PR #28, opened 2017-05-04
add sources and javadoc artifacts

malaffoon:
I was getting tired of not having the source for RDW_Common modules. This generates both the sources and javadoc jars. I had to fix a bunch of javadoc to get it to properly generate without warnings.


malaffoon:
Eventually, yeah. But it's not quite as important for apps since they aren't pulled into other projects.

PR #30, opened 2017-06-02
move AssessmentType and Subject into rdw common

malaffoon:
* rename package from `rdw.model` to `rdw.common.model`
* push TRT-related classes down into `rdw.common.model.trt`
* add enums

Obviously, i'll have a bunch of changes in ingest and reporting coming up.


model/src/main/java/org/opentestsystem/rdw/common/model/AssessmentType.java
```diff
@@ -0,0 +1,56 @@
...
+ * The 'code' corresponds to the database id (don't change the codes arbitrarily!).
+ * 'hasClaims' is false for IABs, true for others.
+ * </p>
+ */
+public enum AssessmentType {
+    ICA(1, true),
+    IAB(2, false),
+    SUMMATIVE(3, true);
+
+    private final int code;
```
malaffoon: I renamed this from `value` to `code` because value has connotations for enums.

model/src/main/java/org/opentestsystem/rdw/common/model/AssessmentType.java
```diff
@@ -0,0 +1,56 @@
...
+ * The 'id' corresponds to the database id (don't change the ids arbitrarily!).
+ * 'hasClaims' is false for IABs, true for others.
+ * </p>
+ */
+public enum AssessmentType {
+    ICA(1, true),
+    IAB(2, false),
+    SUMMATIVE(3, true);
+
+    private final int id;
```
malaffoon: I renamed this from `value` to `id` because value has connotations with enums.

PR #32, opened 2017-06-06
modifying build version string to be Maven standard compliant

rcordeau:


build.gradle
```diff
@@ -23,7 +23,7 @@ allprojects {
     apply plugin: 'propdeps-idea'
 
     if (project.hasProperty("buildNumber")) {
-      project.setVersion(project.version.split("-")[0] + "." + buildNumber)
+      project.setVersion(project.version.toString().replaceAll("SNAPSHOT", "${buildNumber}"))
```
malaffoon: 👍   much safer choice

PR #33, opened 2017-08-04
Add InputStream read access to the ArchiveService.

wtritch:
This PR just adds the ability to read an unbuffered InputStream from the ArchiveService.  The goal is to be able to stream a multi-hundred-megabyte report PDF from S3 to the user without copying the whole file to memory.

### DIFF
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
@@ -37,6 +37,14 @@ public interface ArchiveService {
     byte[] readResource(String location);
 
     /**
+     * Read a resource identified by the unique id.
+     *
+     * @param location unique (relative) location for content, e.g. "EXAM/4F/0A/4F0A1245"
+     * @return a content stream, null if not found
+     */
+    InputStream readResourceAsStream(String location);
+
+    /**
      * Read the properties for the given resource location.
      *
      * @param location unique (relative) location for content
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
@@ -87,8 +87,21 @@ public class S3ArchiveService implements ArchiveService, InitializingBean {
 
     @Override
     public byte[] readResource(final String location) {
-        try (final InputStream is = new BufferedInputStream(amazonS3.getObject(bucket, location).getObjectContent())) {
-            return StreamUtils.copyToByteArray(is);
+        try (final InputStream is = readResourceAsStream(location)) {
+            if (is == null) return null;
+
+            return StreamUtils.copyToByteArray(new BufferedInputStream(is));
+        } catch (final IOException e) {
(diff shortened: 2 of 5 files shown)


archive/src/test/java/org/opentestsystem/rdw/archive/S3ArchiveServiceIT.java
```diff
@@ -63,4 +66,19 @@ public void itShouldWriteAndReadContentInLocation() throws UnsupportedEncodingEx
         assertThat(readProperties.getProperty("content-type")).isEqualTo("text/plain");
         assertThat(readProperties.getProperty("message")).isEqualTo("hi");
     }
+
+    @Ignore("need credentials")
+    @Test
+    public void itShouldWriteAndReadContentUsingStreams() throws Exception {
```
wtritch: I put my credentials in this class and successfully ran all the tests, double-checking that the files were created in S3 and cleaning up after myself afterwards. =)

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -87,19 +87,28 @@ public void writeResource(final String location, final InputStream is, final Pro
 
     @Override
     public byte[] readResource(final String location) {
-        try (final InputStream is = new BufferedInputStream(amazonS3.getObject(bucket, location).getObjectContent())) {
-            return StreamUtils.copyToByteArray(is);
+        try (final InputStream is = readResourceAsStream(location)) {
```
wtritch: delegate to opening the stream, then just read it into a byte array.

chdietz:
Do we plan on expiring these generated reports after some time?  If so not sure if that has to be at the bucket level or if it can be in a folder.  Also wondering should the batch pdf files be in the same bucket as ingest content?

wtritch:
Yeah, the goal is expiration with the ability to regenerate expired reports using current permissions.  I'm thinking a separate bucket would be ideal, to give us more granular permission control, but that can be configured on a per-Deployment basis

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
```diff
@@ -37,6 +37,14 @@
     byte[] readResource(String location);
 
     /**
+     * Read a resource identified by the unique id.
+     *
+     * @param location unique (relative) location for content, e.g. "EXAM/4F/0A/4F0A1245"
+     * @return a content stream, null if not found
+     */
+    InputStream readResourceAsStream(String location);
```
malaffoon: perhaps `openResource`?

malaffoon:
If we include properties when we write the report we can purge based on those.

wtritch:
@malaffoon Since purging will also require updating the associated DB record, we can just use the user_report.created field and user_report.report_resource_uri rather than property scanning.

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
wtritch: I like that idea, changed.

PR #36, opened 2017-08-08
remove non-string properties before calling Properties.store

wtritch:
I was running into an issue using the LocalArchiveService when passing in a <"Content-Length", 123L> property.  Properties.store() calls (String) entry.getValue() under the covers, and if it contains any non-string values an exception is thrown.

This PR just strips non-String entries from the Properties object before persisting.


archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -72,6 +72,10 @@ public void writeResource(final String location, final InputStream is, final Pro
             return;
         }
 
+        //Remove non-String property values prior to storing
+        properties.entrySet()
+                .removeIf(entry -> !String.class.isInstance(entry.getValue()));
```
esotrope: So I know I'm missing some context here probably but I'm guessing we are storing some metadata in a .properties file next to our archived report?
Was this bug a side effect of how the old java Properties class stores/loads properties from a file without automagic type conversion?

malaffoon: Yes, we are storing properties in a properties file. And, yes, the dumb Properties class is partly to blame; this started happening because we started storing content-length which we hadn't before. 

In retrospect i should probably have used XML storage of the properties but ... well, ugh.

esotrope: I'm confused, so when we write the file to S3, the S3 FS doesn't store the file length as a property on the inode or whatever they use on their FS?

malaffoon: Oh, this is only in the LocalArchiveService which is writing to local file system. No S3 involved. This is really only a test / local-dev shim.

esotrope: Ah got it - thanks.

PR #39, opened 2017-08-10
refactor bean creation for ArchiveService

malaffoon:
The way it was before, it was not possible to instantiate multiple ArchiveService's because it was relying on the Spring auto-configured amazon S3 client. It also didn't allow dynamic creation of the ArchiveService beans. 

This is a problem for production because ETS and ITS both want copies of the recon reports delivered to them, to two different buckets. Plus even a single archive report sender was not getting wired up properly to the amazon S3 client. 

Anyway, this switches to a factory-based approach.

I'll be making changes to the ingest apps, import-service and task-service; as well as the report-processor app.


archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveConfiguration.java
```diff
@@ -1,49 +1,92 @@
...
-     * The root is expected to be of the form "file:///tmp/" or "s3://bucket"
-     *
-     * @param archiveRoot configured archive root
-     * @return new created {@link ArchiveService}
-     * @throws IllegalArgumentException if root is invalid or not found
-     */
     @Bean
-    public ArchiveService archiveService(@Value("${archive.root:file:///tmp/}") final String archiveRoot) {
-        return ArchiveServiceFactory.newInstance(archiveRoot);
+    public ArchiveServiceFactory archiveServiceFactory() {
```
malaffoon: a lot of this is due to the relaxed property resolution ... bit messier but definitely worth it.

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -32,35 +30,28 @@
...
 
-    private final String bucket;
     private AmazonS3 amazonS3;
+    private final String bucket;
 
     /**
      * @param root root path, must start with "s3://", e.g. "s3://myBucket"
      */
-    S3ArchiveService(final String root) {
+    S3ArchiveService(final AmazonS3 amazonS3, final String root) {
```
malaffoon: This is now a lot cleaner (using constructor instead of setter and afterPropertiesSet hook); makes unit testing easier too.

chdietz:
@malaffoon, if I am using archive service and using yml Archive: root: s3://bucket do I have to make any changes?

malaffoon:
@chdietz yes probably a slight change. Instead of autowiring an ArchiveService, you'll autowire an ArchiveServiceFactory and generate the archive service bean yourself. My ingest PR (coming soon) will demonstrate that.

chdietz:
What about two buckets with different credentials?

malaffoon:
Two buckets with different credentials worked fine in my testing (that was kind of the point of this change).

chdietz:
cool!

PR #40, opened 2017-08-11
add option to specify SSE (server side encryption algorithm)

malaffoon:
ETS is using SSE for their S3 bucket so we have to be able to specify it in the configuration. Surprisingly (well, not really) the AmazonS3Client doesn't allow this to be set so we have to inject it ourselves with each put request.


archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -105,7 +115,7 @@ public void writeResource(final String location, final InputStream is, final Pro
         try {
             amazonS3.putObject(bucket, location, content, metadata);
         } catch (final AmazonServiceException e) {
-            final String msg = "Error writing content to " + bucket + location;
+            final String msg = "Error writing content to " + bucket + "/" + location;
```
malaffoon: unrelated, the error message was driving me nuts without the slash

archive/src/test/java/org/opentestsystem/rdw/archive/MultiS3ArchiveServiceIT.java
```diff
@@ -0,0 +1,44 @@
+package org.opentestsystem.rdw.archive;
```
malaffoon: This was me chasing another theory about how two amazon clients were interfering with each other.
They weren't.
I successfully had 4 clients writing to different buckets in different regions with different credentials.
But i hate to delete test code.

malaffoon:
I'm going to merge this so i can move forward with ETS testing.

PR #41, opened 2017-08-16
catch client exception

malaffoon:
See DWR-756 for why this change.


security/src/main/java/org/opentestsystem/rdw/security/client/permissionservice/PermissionWebServiceClient.java
```diff
@@ -36,7 +40,7 @@ public PermissionWebServiceClient(@NotNull final RestTemplate restTemplate, @Not
 	 * @return response containing all registered components
 	 */
 	public Response<Component> getComponents() {
-		return get(endpoint + "/component", new ParameterizedTypeReference<Response<Component>>() {});
+		return get("/component", Component.class);
```
malaffoon: Slightly unrelated, just further reducing duplication: every method was repeating these things.

jtreuting:
Nice addition

PR #42, opened 2017-08-16
partially revert changes (it was losing type information somehow)

malaffoon:


security/src/main/java/org/opentestsystem/rdw/security/client/permissionservice/PermissionWebServiceClient.java
```diff
@@ -40,7 +40,9 @@ public PermissionWebServiceClient(@NotNull final RestTemplate restTemplate, @Not
 	 * @return response containing all registered components
 	 */
 	public Response<Component> getComponents() {
-		return get("/component", Component.class);
+		return get("/component",
+                   new ParameterizedTypeReference<Response<Component>>() {},
```
malaffoon: I think this was the issue ... it wasn't inferring response type or something.

security/src/main/java/org/opentestsystem/rdw/security/client/permissionservice/PermissionWebServiceClient.java
```diff
@@ -78,24 +86,18 @@ public PermissionWebServiceClient(@NotNull final RestTemplate restTemplate, @Not
...
-    private <T> Response<T> get(String path, Class<T> type, Map<String, ?> parameters) {
+    private <T> T get(String path, ParameterizedTypeReference<T> typeReference, Map<String, ?> parameters) {
+        final String url = endpoint + path;
         try {
-            return restTemplate.exchange(endpoint + path, HttpMethod.GET, null,
-                    new ParameterizedTypeReference<Response<T>>() {}, parameters).getBody();
+            return restTemplate.exchange(url, HttpMethod.GET, null, typeReference, parameters).getBody();
         } catch (final RestClientException e) {
-            logger.warn("Error getting {} from {}{}", type.getName(), endpoint + path, parameters.isEmpty() ? "" : (" component=" + parameters.get("component")));
+            logger.warn("Error getting {}{}: {}", url, parameters.isEmpty() ? "" : (" component=" + parameters.get("component")), e.getMessage());
```
esotrope: Should this really log or should it just throw a runtime exception and let the client log it?

malaffoon: Perhaps. But the way it is now the client is just checking for a null response and not catching the exception. So it eventually gets logged as an error in response to what the client is trying to do ... and, out of context, it is very misleading. Since the contract on this service appears to be it-will-return-null-on-error, i took this approach. We could, of course, change the contract and catch the exception in the client.

PR #43, opened 2017-08-24
adding ability to get the archived file's uri root from the saved properties

rcordeau:
I don't really like it much, but it works for now.

### DIFF
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
@@ -76,6 +76,7 @@ public class LocalArchiveService implements ArchiveService {
 
         try (final Writer writer = new BufferedWriter(new FileWriter(realPropertiesFile(location), false))) {
             normalizeProperties(properties).store(writer, null);
+            addRootURI(properties).store(writer, null);
         } catch (final IOException e) {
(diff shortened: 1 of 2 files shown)


archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -76,6 +76,7 @@ public void writeResource(final String location, final InputStream is, final Pro
 
         try (final Writer writer = new BufferedWriter(new FileWriter(realPropertiesFile(location), false))) {
             normalizeProperties(properties).store(writer, null);
+            addRootURI(properties).store(writer, null);
```
malaffoon: I don't like this approach. I would set the "rootURI" in the properties only on the way out, same as content length.

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -207,4 +208,12 @@ private static Properties normalizeProperties(final Properties properties) {
 
         return properties;
     }
+
+    private Properties addRootURI(final Properties properties) {
+        if (properties == null) { return properties; }
+
+        properties.setProperty("rootURI", rootPath);
```
malaffoon: I'm curious why you choose to do the "rootURI" instead of the URI of the actual archive item. Seems like what it should return is the URI for the item ... that's really the thing we need, isn't it?

PR #44, opened 2017-08-24
fixing unit tests

rcordeau:


wtritch:
We should also update the S3ArchiveService tests.  I know they're all ignored during check-in to avoid bleeding credentials, but they should run locally with creds when not ignored.

rcordeau:
This is all changed now.

PR #45, opened 2017-08-24
change the way rawURI property is handled

malaffoon:
Since i was the one whining about it, here is the change i'd like ...


archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -135,8 +134,10 @@ public Properties readProperties(final String location) {
         // cleanup legacy properties
         normalizeProperties(properties);
 
-        // add the content length
-        properties.put(Headers.CONTENT_LENGTH, realResourceFile(location).length());
+        // add the content length and raw URI
+        final File realFile = realResourceFile(location);
+        properties.put(Headers.CONTENT_LENGTH, realFile.length());
+        properties.put(ArchiveService.RawURI, FileSchemePrefix + realFile.getAbsolutePath());
```
malaffoon: only add the property on the way out

archive/src/test/java/org/opentestsystem/rdw/archive/S3ArchiveServiceIT.java
```diff
@@ -48,9 +50,10 @@ public void itShouldWriteAndReadContent() throws UnsupportedEncodingException {
         assertThat(service.exists("TEST/test")).isTrue();
         assertThat(service.readResource("TEST/test")).isEqualTo(content);
         final Properties readProperties = service.readProperties("TEST/test");
-        assertThat(readProperties).hasSize(2);
+        assertThat(readProperties).hasSize(3);
         assertThat(readProperties.get("Content-Length")).isEqualTo(15L);
         assertThat(readProperties.getProperty("Content-Type")).isEqualTo("application/octet-stream");
+        assertThat(readProperties.getProperty(ArchiveService.RawURI)).isEqualTo(root + "/TEST/test");
```
malaffoon: look: i changed the test without checking in my credentials ... that's, like, twice in a row now.

rcordeau:
yes, much cleaner than my approach.

archive/src/test/java/org/opentestsystem/rdw/archive/S3ArchiveServiceIT.java
rcordeau: Yes, I see.

PR #46, opened 2017-09-01
Register the StatusEndpoint as a bean for easy configuration with oth…

wtritch:
…er actuator endpoints.


wtritch:
Merging for speed, will resolve issues as they come up.

PR #48, opened 2017-09-24
refactor Grant to handle district/school groups, and remove duplication

malaffoon:
I removed the security.Grant class using the one from utils instead. Just trying to reduce duplication.

It seems like much of the Grant functionality in common is no longer used. For example, i spent some time getting the "contains" algorithm more correct only to discover that it isn't used anywhere. In retrospect that makes sense, since the de-duplication now happens a the permission level instead of the role level.

I also pulled over some code that was common to admin and reporting. Not quite ready to move `AuthorizationService` though.

Anyway, i have coded the changes in ingest and reporting, and everything seems to work. Those PR's will be forthcoming.


build.gradle
```diff
@@ -15,7 +15,7 @@ apply plugin: 'io.spring.dependency-management'
 
 allprojects {
     group = 'org.opentestsystem.rdw.common'
-    version = '1.0.0-SNAPSHOT'
+    version = '1.1.0-SNAPSHOT'
```
malaffoon: My ongoing quest to bump everything to 1.1.0

utils/src/main/java/org/opentestsystem/rdw/utils/Grant.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Objects;
+
+import static org.opentestsystem.rdw.utils.Grant.Level.CLIENT;
+import static org.opentestsystem.rdw.utils.Grant.Level.STATE;
+
+/**
+ * A helper class that wraps the field tokens providing named accessors.
+ * It also does basic validation between grant level and field data.
+ */
+public class Grant {
```
malaffoon: This used to be inside TenancyChain; pulled it out to be used in reporting.

utils/src/main/java/org/opentestsystem/rdw/utils/Grant.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        return new Grant(tokens);
+    }
+
+    public String toString() {
+        return String.join("|", tokens);
+    }
+
+    public boolean similar(final Grant that) {
```
malaffoon: The security.Grant class had an equals method that looked like this. I chose to make the equals here more rigorous and add this helper instead. It seems to work fine in all places it is used, even the Set<Grant> in reporting. But i could be missing some subtlety for certain data sets?

utils/src/main/java/org/opentestsystem/rdw/utils/Grant.java
```diff
@@ -0,0 +1,147 @@
...
+     * Determine if this grant "contains" another grant. This is true if the grants refer to the
+     * same role and this grant is at the same or higher level (e.g. DISTRICT is higher than
+     * INSTITUTION) and has the same ancestry of the other grant. For example, if this grant is
+     * role A for district D1, and that grant is role A for school S1 in district D1, then this
+     * grant contains that grant.
+     *
+     * @param that the grant to check
+     * @return true if this grant contains that grant
+     */
+    public boolean contains(final Grant that) {
```
malaffoon: There were logical flaws in the security.Grants contains method that i tried to fix here. Then i realized it wasn't being used anywhere, but i couldn't bring myself to remove this method.

utils/src/test/java/org/opentestsystem/rdw/utils/GrantTest.java
```diff
@@ -0,0 +1,96 @@
...
+        assertThat(g1.contains(g2)).isTrue();
+        assertThat(g1.contains(g3)).isTrue();
+        assertThat(g1.contains(g5)).isTrue();
+        assertThat(g1.contains(g6)).isTrue();
+        assertThat(g2.contains(g5)).isTrue();
+        assertThat(g2.contains(g6)).isTrue();
+
+        assertThat(g1.contains(g4)).isFalse();  // mismatch role
+        assertThat(g2.contains(g3)).isFalse();  // mismatch entity
+        assertThat(g3.contains(g5)).isFalse();  // mismatch district
```
malaffoon: `g3.contains(g5)` is an example of the logic flaw in the original algorithm (which returned true).

build.gradle
wtritch: does every external project need to go bump its dependency?

utils/src/main/java/org/opentestsystem/rdw/utils/Grant.java
wtritch: maybe a javadoc?  "similar" feels pretty vague to me.  "equivalent" "compatibleWith" ?

security/src/main/java/org/opentestsystem/rdw/security/PermissionScope.java
```diff
@@ -0,0 +1,140 @@
...
+            final Collection<Long> institutionGroupIds,
+            final Collection<Long> institutionIds) {
+        this.districtGroupIds = districtGroupIds != null ? ImmutableSet.copyOf(districtGroupIds) : ImmutableSet.of();
+        this.districtIds = districtIds != null ? ImmutableSet.copyOf(districtIds) : ImmutableSet.of();
+        this.institutionGroupIds = institutionGroupIds != null ? ImmutableSet.copyOf(institutionGroupIds) : ImmutableSet.of();
+        this.institutionIds = institutionIds != null ? ImmutableSet.copyOf(institutionIds) : ImmutableSet.of();
+        this.statewide = false;
+    }
+
+    public boolean isStatewide() {
```
wtritch: I know you didn't write this, but maybe some javadocs will help the next developer trying to grok the permissions system?

utils/src/main/java/org/opentestsystem/rdw/utils/TenancyChain.java
```diff
@@ -141,9 +149,18 @@ public boolean hasRoleForState(final String role, final String state) {
      * @return true if role is granted for given district
      */
     public boolean hasRoleForDistrict(final String role, final String district) {
-        return grants.stream().anyMatch(grant -> "DISTRICT".equalsIgnoreCase(grant.getLevel())
-                && role.equalsIgnoreCase(grant.getRole())
-                && (district.equalsIgnoreCase(grant.getDistrictId()) || district.equalsIgnoreCase(grant.getDistrictName())));
+        return grants.stream().anyMatch(grant -> grant.hasRoleForLevel(role, DISTRICT, district));
+    }
```
wtritch: no similar method for DISTRICT_GROUP ?

utils/src/main/java/org/opentestsystem/rdw/utils/Grant.java
esotrope: If the tenancy chain obeys its contract this is a valid equals() method.

utils/src/main/java/org/opentestsystem/rdw/utils/Grant.java
```diff
@@ -0,0 +1,147 @@
...
+        CLIENT(4),
+        STATE_GROUP(6),
+        STATE(8),
+        DISTRICT_GROUP(10),
+        DISTRICT(12),
+        INSTITUTION_GROUP(14),
+        INSTITUTION(16);
+
+        // index of id field in tenancy chain
+        private final int idx;
```
esotrope: nit: no need to abbreviate this.

utils/src/main/java/org/opentestsystem/rdw/utils/Grant.java
```diff
@@ -0,0 +1,147 @@
...
+        int idx() {
+            return this.idx;
+        }
+    }
+
+    private static final int idxEntityId = 1;
+    private static final int idxRole = 2;
+    private static final int idxLevel = 3;
+
+    private final String[] tokens;
```
esotrope: What is the reason for maintaining the tenancy chain contents as an array? Why not parse out what all the fields at each index are into named fields? Do we iterate over the chain to for a method?

build.gradle
malaffoon: My goal is simply that the develop branch for all the RDW projects reflects the next release, 1.1.x.

utils/src/main/java/org/opentestsystem/rdw/utils/TenancyChain.java
malaffoon: Yeah, these aren't actually used any more. I'll probably clean up a couple of them (only the client and state ones really make sense at this point).

utils/src/main/java/org/opentestsystem/rdw/utils/Grant.java
malaffoon: will fix

utils/src/main/java/org/opentestsystem/rdw/utils/Grant.java
malaffoon: I guess it is left over from the initial implementation: there didn't seem to be a reason to copy the fields into named variables since we already have the values in string form.

security/src/main/java/org/opentestsystem/rdw/security/PermissionScope.java
malaffoon: I added some class-level javadocs

PR #49, opened 2017-09-26
add tests for serialization ... and make them pass

malaffoon:
I'm going to merge this so i can push forward in reporting and admin. I will, of course, address any issues brought up here ...


security/src/main/java/org/opentestsystem/rdw/security/PermissionScope.java
```diff
@@ -35,6 +35,13 @@
     private final Set<Long> institutionIds;
     private final boolean statewide;
 
+    /**
+     * Default constructor for serialization
+     */
+    private PermissionScope() {
+       this(false);
+    }
```
malaffoon: This was the missing bit. It was a copy&paste error on my part.

PR #50, opened 2017-10-09
Updated LocalArchiveService to always return properties even when a p…

bgsavage:
…roperties file is not found.

### DIFF
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
@@ -40,8 +40,8 @@ public class LocalArchiveService implements ArchiveService {
     private final String rootPath;
 
     public static final Map<String, String> ExtensionToMediaType = ImmutableMap.<String, String>builder()
-        .put("csv", "text/plain")
-        .build();
+            .put("csv", "text/plain")
+            .build();
 
     /**
      * @param root root path, must start with "file://", e.g. "file:///tmp/"
@@ -148,7 +148,7 @@ public class LocalArchiveService implements ArchiveService {
         // If the content type is not set, guess it from the file name.
         if (!properties.containsKey(Headers.CONTENT_TYPE)) {
             String contentType = guessContentType(resourceFile.getName());
-            if(contentType != null) {
+            if (contentType != null) {
                 properties.put(Headers.CONTENT_TYPE, contentType);
             }
         }
@@ -225,13 +225,14 @@ public class LocalArchiveService implements ArchiveService {
     }
 
     private static String guessContentType(final String filename) {
-        String contentType = URLConnection.guessContentTypeFromName(filename);
-        if(contentType == null) {
-            String extension = Files.getFileExtension(filename).toLowerCase();
-            if(ExtensionToMediaType.containsKey(extension))
-                contentType = ExtensionToMediaType.get(extension);
-        }
+        final String contentTypeFromName = URLConnection.guessContentTypeFromName(filename);
+        if (contentTypeFromName != null)
+            return contentTypeFromName;
 
-        return contentType;
+        String extension = Files.getFileExtension(filename).toLowerCase();
+        if (ExtensionToMediaType.containsKey(extension))
+            return ExtensionToMediaType.get(extension);
+
+        return null;
     }
 }
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
@@ -1,6 +1,8 @@
 package org.opentestsystem.rdw.archive;
 
 import com.amazonaws.services.s3.Headers;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.io.Files;
 import com.google.common.io.MoreFiles;
 import com.google.common.io.RecursiveDeleteOption;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -209,4 +223,15 @@ private static Properties normalizeProperties(final Properties properties) {
...
         return properties;
     }
+
+    private static String guessContentType(final String filename) {
+        String contentType = URLConnection.guessContentTypeFromName(filename);
+        if(contentType == null) {
+            String extension = Files.getFileExtension(filename).toLowerCase();
+            if(ExtensionToMediaType.containsKey(extension))
+                contentType = ExtensionToMediaType.get(extension);
+        }
```
bgsavage: First I'm trying to use the URLConnection method, and if not, then then look in the hardcoded map.  Not too happy with the hardcoded map, but hopefully it doesn't grow too much.

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -117,27 +125,33 @@ public Properties readProperties(final String location) {
...
         }
 
         // cleanup legacy properties
         normalizeProperties(properties);
 
         // add the content length and raw URI
-        final File realFile = realResourceFile(location);
-        properties.put(Headers.CONTENT_LENGTH, realFile.length());
-        properties.put(ArchiveService.RawURI, FileSchemePrefix + realFile.getAbsolutePath());
+        final File resourceFile = realResourceFile(location);
```
bgsavage: unrelated: I renamed `file` and `realFile` for clarity

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -209,4 +223,15 @@ private static Properties normalizeProperties(final Properties properties) {
 
         return properties;
     }
+
+    private static String guessContentType(final String filename) {
```
esotrope: Would prefer to see the method structured like below.
This way we don't update the same variable as we go and we dont have to track the state of that variable as we read the method.

```java
private static String guessContentType(final String filename) {
  final String firstGuess = URLConnection.guessContentTypeFromName(filename);
  if (firstGuess != null) {
    return firstGuess;
  }
  final String secondGuess = ExtensionToMediaType.get(Files.getFileExtension(filename).toLowerCase());
  if (secondGuess != null) {
    return secondGuess;
  }
  return null;
}
```

archive/src/test/java/org/opentestsystem/rdw/archive/LocalArchiveServiceTest.java
```diff
@@ -1,5 +1,6 @@
 package org.opentestsystem.rdw.archive;
 
+import com.amazonaws.services.s3.Headers;
```
esotrope: I think it would be more portable to use spring's HttpHeaders class here

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
bgsavage: Alright alright, >nit: final , :P

Thanks for the suggestion, I'll update.

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -117,27 +125,33 @@ public Properties readProperties(final String location) {
...
-            final String msg = "Error reading properties from " + file.getPath();
-            logger.warn(msg, e);
-            throw new RuntimeException(msg, e);
+        final File propertiesFile = realPropertiesFile(location);
+
+        if (propertiesFile.isFile()) {
+            try (final Reader reader = new BufferedReader(new FileReader(propertiesFile))) {
+                properties.load(reader);
+            } catch (final IOException e) {
+                final String msg = "Error reading properties from " + propertiesFile.getPath();
```
esotrope: nit: abbr

wtritch:
Is this so that you can rely on CONTENT_TYPE and CONTENT_LENGTH always being present in local development?

archive/src/test/java/org/opentestsystem/rdw/archive/LocalArchiveServiceTest.java
```diff
@@ -117,4 +117,16 @@ public void itShouldDelete() {
...
+    @Test
+    public void itShouldReadPropertiesWhenPropertiesFileDoesNotExit() {
+        final byte[] content = "test1".getBytes();
+
+        service.writeResource("testNoProps.csv",  content, null);
+
+        Properties properties = service.readProperties("testNoProps.csv");
+
+        assertThat(properties.get(Headers.CONTENT_LENGTH)).isEqualTo(5L);
+        assertThat(properties.get(Headers.CONTENT_TYPE)).isEqualTo("text/plain");
```
malaffoon: why didn't that return "text/csv" or "application/csv"?

archive/src/test/java/org/opentestsystem/rdw/archive/LocalArchiveServiceTest.java
malaffoon: Perhaps. But the test and the impl should use the same import (and right now, the impl uses this import)

bgsavage:
@wtritch Yes

archive/src/test/java/org/opentestsystem/rdw/archive/LocalArchiveServiceTest.java
bgsavage: To have consistency in the content-type returned by both S3 and Local.  "text/plan" is just a broader umbrella, no?

PR #51, opened 2017-10-10
Feature/add get raw uri

bgsavage:
Added the `getRawUri` method to ArchiveService and related tests.

### DIFF
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
@@ -16,6 +16,13 @@ public interface ArchiveService {
     String RawURI = "rawURI";
 
     /**
+     * Gets the Archive Service fully qualified URI for a given location.
+     * @param location unique (relative) location for content, e.g. "EXAM/4F/0A/4F0A1245"
+     * @return the raw uri e.g. "file:///tmp/EXAM/4F/0A/4F0A1245"
+     */
+    String getRawUri(String location);
+
+    /**
      * Byte array version of writeResource; intended for small resources; retained for legacy use.
      *
      * @param location unique (relative) location for content, e.g. "EXAM/4F/0A/4F0A1245"
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
@@ -16,7 +16,8 @@ public interface ArchiveService {
     String RawURI = "rawURI";
 
     /**
-     * Gets the Archive Service fully qualified URI for a given location.
+     * Gets the Archive Service fully qualified URI for a given location, regardless of whether
+     * the resource exists or not.
      * @param location unique (relative) location for content, e.g. "EXAM/4F/0A/4F0A1245"
      * @return the raw uri e.g. "file:///tmp/EXAM/4F/0A/4F0A1245"
      */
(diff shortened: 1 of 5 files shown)


archive/src/test/java/org/opentestsystem/rdw/archive/LocalArchiveServiceTest.java
```diff
@@ -142,4 +142,25 @@ public void itShouldGuessContentTypeSometimes() {
         assertThat(LocalArchiveService.guessContentType("test.xls")).isNull();
         assertThat(LocalArchiveService.guessContentType("file:///test.txt")).isEqualTo("text/plain");
```
bgsavage: Thanks @malaffoon for adding these tests.

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -115,9 +120,9 @@ public void writeResource(final String location, final InputStream is, final Pro
         try {
             amazonS3.putObject(bucket, location, content, metadata);
         } catch (final AmazonServiceException e) {
-            final String msg = "Error writing content to " + bucket + "/" + location;
-            logger.warn(msg, e);
-            throw new RuntimeException(msg, e);
+            final String message = "Error writing content to " + getRawUri(location);
+            logger.warn(message, e);
+            throw new RuntimeException(message, e);
```
bgsavage: unrelated rename.

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
```diff
@@ -16,6 +16,13 @@
     String RawURI = "rawURI";
 
     /**
+     * Gets the Archive Service fully qualified URI for a given location.
+     * @param location unique (relative) location for content, e.g. "EXAM/4F/0A/4F0A1245"
```
malaffoon: Perhaps add a clause like `regardless of whether a resource exists for the given location`.

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -55,6 +55,11 @@ public LocalArchiveService(final String root) {
     }
 
     @Override
+    public String getRawUri(final String location) {
+        return FileSchemePrefix + rootPath + location;
```
malaffoon: I think you might want to go through the realResourceFile() method. All that that fiddling with getCanonicalFile is there because of symlinks and stuff. I think the most correct form of this might be: `return FileSchemePrefix + realResourceFile(location).getAbsolutePath();`

archive/src/test/java/org/opentestsystem/rdw/archive/LocalArchiveServiceTest.java
```diff
@@ -142,4 +142,25 @@ public void itShouldGuessContentTypeSometimes() {
...
     }
+    
+    @Test
+    public void itShouldReturnUriForALocation(){
+        String location = "some/location/file.pdf";
```
malaffoon: nit: `final`

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
bgsavage: 👍

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
bgsavage: Hm, so we want to convert symlinks to absolute paths?  I'm not doing this with the export-root in ArchiveExamExportService... If we should be doing it there, than that's only specific to a local implementation, which would warrant separate implementations again.

PR #52, opened 2017-10-11
Added `writeProperties` function to ArchiveService to allow for prope…

bgsavage:
…rties to be written /  updated after a resource has been stored.

### DIFF
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
@@ -40,7 +40,7 @@ public class LocalArchiveService implements ArchiveService {
     private final String rootPath;
 
     private static final Map<String, String> ExtensionToMediaType = ImmutableMap.<String, String>builder()
-            .put("csv", "text/plain")
+            .put("csv", "text/csv")
             .build();
 
     /**
@@ -83,18 +83,9 @@ public class LocalArchiveService implements ArchiveService {
             throw new RuntimeException(msg, e);
         }
 
-        if (properties == null || properties.isEmpty()) {
-            return;
+        if (properties != null && !properties.isEmpty()) {
+            writePropertiesToFile(location, properties);
         }
-
-        try (final Writer writer = new BufferedWriter(new FileWriter(realPropertiesFile(location), false))) {
-            normalizeProperties(properties).store(writer, null);
-        } catch (final IOException e) {
-            final String msg = "Error writing properties to " + realPropertiesFile(location).getPath();
-            logger.warn(msg, e);
-            throw new RuntimeException(msg, e);
-        }
-
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -40,7 +40,7 @@
     private final String rootPath;
 
     private static final Map<String, String> ExtensionToMediaType = ImmutableMap.<String, String>builder()
-            .put("csv", "text/plain")
+            .put("csv", "text/csv")
```
bgsavage: Now that we can set content-type and not rely on what s3 returns by default, using more refined `text/csv`

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -74,24 +75,8 @@ public void writeResource(final String location, final byte[] content, Propertie
...
-                } else if (Headers.CONTENT_LENGTH.equalsIgnoreCase(key)) {
-                    metadata.setContentLength((Long) entry.getValue());
-                    lengthKnown = true;
-                } else {
-                    metadata.addUserMetadata(key, entry.getValue().toString());
-                }
-            }
-        }
+        final ObjectMetadata metadata = mapPropertiesToObjectMetadata(properties);
+        final boolean lengthKnown = (metadata.getContentLength() > 0);
```
bgsavage: Now `lengthKnown` can be `final` since it's not being set inside the for loop.

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -217,4 +213,27 @@ static boolean validRoot(final String root) {
...
+                final String key = (String) entry.getKey();
+
+                // extract S3-specific properties into appropriate metadata
+                // and all other properties into user metadata
+                if (Headers.CONTENT_TYPE.equalsIgnoreCase(key)) {
+                    metadata.setContentType((String) entry.getValue());
+                } else if (Headers.CONTENT_LENGTH.equalsIgnoreCase(key)) {
+                    metadata.setContentLength((Long) entry.getValue());
+                } else {
+                    metadata.addUserMetadata(key, entry.getValue().toString());
```
bgsavage: Per the java docs of this method, it will create a new key if it doesn't exist, or overwrite if it does.

archive/src/test/java/org/opentestsystem/rdw/archive/LocalArchiveServiceTest.java
```diff
@@ -127,13 +166,12 @@ public void itShouldReadPropertiesWhenPropertiesFileDoesNotExit() {
         Properties properties = service.readProperties("testNoProps.csv");
 
         assertThat(properties.get(Headers.CONTENT_LENGTH)).isEqualTo(5L);
-        assertThat(properties.get(Headers.CONTENT_TYPE)).isEqualTo("text/plain");
+        assertThat(properties.get(Headers.CONTENT_TYPE)).isEqualTo("text/csv");
     }
 
     @Test
     public void itShouldGuessContentTypeSometimes() {
-        // these are the ones we're likely to have in the archive (should we fix *.csv?)
```
bgsavage: Yes.  :)

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -162,6 +135,17 @@ public Properties readProperties(final String location) {
     }
 
     @Override
+    public void writeProperties(final String location, final Properties properties) {
+        final Properties fileProperties = readPropertiesFromFile(location);
```
malaffoon: So this is additive? There is no way to clear a property. Did you think about making it replace the existing (user) properties?

bgsavage: I don't think a replace would be safe in that we add some props that the consumer is never aware of (i.e.`sse`).

wtritch: hmm... maybe we could interpret a "null" value as a "clear" ?

bgsavage: Can we decipher the `null` difference of the property no longer existing vs. explicitly set as `null`?

malaffoon: Properties can't contain null values can they?

PR #53, opened 2017-10-19
DWR-1092  Change getEffectiveDate to adhere to the Data Model format …

agorina:
…(which is 'YYYY-MM-DD').

### DIFF
--- a/model/src/main/java/org/opentestsystem/rdw/common/model/trt/Opportunity.java
+++ b/model/src/main/java/org/opentestsystem/rdw/common/model/trt/Opportunity.java
@@ -1,11 +1,5 @@
 package org.opentestsystem.rdw.common.model.trt;
 
-import java.time.Instant;
-import java.time.LocalDate;
-import java.time.ZoneId;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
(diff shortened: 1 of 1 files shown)


model/src/main/java/org/opentestsystem/rdw/common/model/trt/Opportunity.java
```diff
@@ -881,15 +883,15 @@ public void setAssessmentParticipantSessionPlatformUserAgent(String value) {
...
      * @return
      *     possible object is
-     *     {@link Date }
+     *     {@link LocalDate }
      *
      */
-    public Date getEffectiveDate() {
-        return effectiveDate;
+    public LocalDate getEffectiveDate() {
+        return effectiveDate.toInstant().atZone(ZoneId.of("GMT")).toLocalDate();
```
agorina: Is this what was the approach recommended in https://github.com/SmarterApp/RDW_Ingest/pull/376?

malaffoon: i want to dig around a little ... if `Date` has a toInstant() then i don't understand why we need to zone it to convert it to `LocalDate`

malaffoon: Did you try `LocalDate.from(effectiveDate.toInstant())`?

chdietz: But Instant is a point in time on the UTC timezone, correct?

agorina: Change it.

PR #55, opened 2017-10-21
DWR-1092  Fix for effectiveDate date.

agorina:


model/src/main/java/org/opentestsystem/rdw/common/model/trt/Opportunity.java
```diff
@@ -890,7 +891,7 @@ public void setAssessmentParticipantSessionPlatformUserAgent(String value) {
      *
      */
     public LocalDate getEffectiveDate() {
-        return effectiveDate == null ? null : LocalDate.from(effectiveDate.toInstant());
+        return effectiveDate == null ? null : effectiveDate.toInstant().atZone(ZoneId.systemDefault()).toLocalDate();
     }
```
agorina: This was giving me a runtime exception. I added a test to show it (should have done before).

model/src/main/java/org/opentestsystem/rdw/common/model/trt/Opportunity.java
```diff
@@ -890,7 +891,7 @@ public void setAssessmentParticipantSessionPlatformUserAgent(String value) {
      *
      */
     public LocalDate getEffectiveDate() {
-        return effectiveDate == null ? null : LocalDate.from(effectiveDate.toInstant());
+        return effectiveDate == null ? null : effectiveDate.toInstant().atZone(ZoneId.systemDefault()).toLocalDate();
```
agorina: I am open to a better solution if anybody has suggestions.

malaffoon: I don't understand this change.

agorina: `java.time.DateTimeException: Unable to obtain LocalDate from TemporalAccessor: 2007-05-02T14:30:00Z of type java.time.Instant

	at java.time.LocalDate.from(LocalDate.java:368)`

malaffoon: Alrighty then, that sucks.

I think the proper solution to this is to convert the date field to a LocalDate during XML parsing. It is at that point that we "know" the expected offset / timezone. Doing it during the get is not appropriate since we don't know how the value was set, so ZoneId.systemDefault() may not produce the correct result. I can dig around a bit but i really think we should change the model to use LocalDate instead of Date.

model/src/test/java/org/opentestsystem/rdw/common/model/trt/OpportunityTest.java
```diff
@@ -19,4 +23,14 @@ public void itShouldReturnDefaultCompleteness() {
...
     }
+
+    @org.junit.Test
+    public void itShouldReturnLocalDateForEffectiveDate() {
+        final Opportunity opportunity = new Opportunity();
+
+        assertThat(opportunity.getEffectiveDate()).isNull();
+
+        opportunity.setEffectiveDate(Date.from(Instant.parse("2007-05-02T14:30:00Z")));
+        assertThat(opportunity.getEffectiveDate()).isEqualTo(LocalDate.of(2007, 5, 2));
```
malaffoon: If the value being parsed is `"2007-05-02T06:30:00Z"` then this test fails. Somehow we need to avoid that.

model/src/main/java/org/opentestsystem/rdw/common/model/trt/Opportunity.java
agorina: I agree with the suggestion to change the model and I can look into it. I am not sure how we would "know" the expected timezone at that time though.

model/src/main/java/org/opentestsystem/rdw/common/model/trt/Opportunity.java
```diff
@@ -123,8 +124,8 @@
     @XmlSchemaType(name = "anySimpleType")
     private String assessmentParticipantSessionPlatformUserAgent;
     @XmlAttribute(name = "effectiveDate", required = true)
-    @XmlJavaTypeAdapter(SimpleDateXmlAdapter.class)
-    private Date effectiveDate;
+    @XmlJavaTypeAdapter(SimpleLocalDateXmlAdapter.class)
```
agorina: Does this look right?

PR #56, opened 2017-10-24
DWR-1119 Exam processor rejects exam from data generator

agorina:


model/src/test/java/org/opentestsystem/rdw/common/model/trt/OpportunityTest.java
```diff
@@ -1,10 +1,6 @@
 package org.opentestsystem.rdw.common.model.trt;
 
 
-import java.time.Instant;
-import java.time.LocalDate;
-import java.util.Date;
```
agorina: Unrelated, just noticed that I left a mess here.

wtritch:
@jtreuting It's called Optional =)

PR #57, opened 2017-12-07
PermissionScope tweaks

malaffoon:
Add copy() for PermissionScope.
Always copy the organization id collections, even if statewide is true.


security/src/main/java/org/opentestsystem/rdw/security/PermissionScope.java
```diff
@@ -152,8 +156,7 @@ public boolean isValid() {
         }
 
         public PermissionScope build() {
-            return statewide ? PermissionScope.STATEWIDE
-                    : new PermissionScope(districtGroupIds, districtIds, schoolGroupIds, schoolIds);
+            return new PermissionScope(statewide, districtGroupIds, districtIds, schoolGroupIds, schoolIds);
```
malaffoon: Here: when doing the embargo stuff, i'm going to need the district ids even if statewide is set true.

security/src/main/java/org/opentestsystem/rdw/security/PermissionScope.java
```diff
@@ -103,6 +98,15 @@ public static Builder builder() {
         private List<Long> schoolGroupIds;
         private List<Long> schoolIds;
 
+        public Builder copy(final PermissionScope scope) {
```
malaffoon: Here: i think i need to copy a user's permissions to remove the statewide flag; otherwise the OrganizationRepo will always return all the districts when statewide=true, and i don't want that.

security/src/main/java/org/opentestsystem/rdw/security/PermissionScope.java
```diff
@@ -103,6 +98,15 @@ public static Builder builder() {
...
+            this.statewide = scope.statewide;
+            this.districtGroupIds = scope.districtGroupIds == null ? null : newArrayList(scope.districtGroupIds);
+            this.districtIds = scope.districtIds == null ? null : newArrayList(scope.districtIds);
+            this.schoolGroupIds = scope.institutionGroupIds == null ? null : newArrayList(scope.institutionGroupIds);
+            this.schoolIds = scope.institutionIds == null ? null : newArrayList(scope.institutionIds);
```
esotrope: nit: inconsistent with other methods using immutables

malaffoon: But it is consistent (in the builder) which uses newArrayList(). I guess i could change those too.

esotrope: ah i think the builder uses null to lazy create the arrays where as the permission scope object itself uses empty collections to be null-safe

security/src/main/java/org/opentestsystem/rdw/security/PermissionScope.java
wtritch: good point

security/src/main/java/org/opentestsystem/rdw/security/PermissionScope.java
wtritch: I remember running into something funky somewhere around here where I couldn't use a Guava Immutable class because nulls were expected... wish I could remember exactly where that was.

PR #59, opened 2018-01-30
add percentile content type

chdietz:
add the NORMS content type to enum

### DIFF
--- a/model/src/main/java/org/opentestsystem/rdw/common/model/ImportContent.java
+++ b/model/src/main/java/org/opentestsystem/rdw/common/model/ImportContent.java
@@ -10,7 +10,8 @@ public enum ImportContent {
     CODES(3),
     ORGANIZATION(4),
     GROUPS(5),
-    EMBARGO(6);
+    EMBARGO(6),
+    PERCENTILE(7);
 
     private final int value;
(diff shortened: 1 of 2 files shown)


model/src/main/java/org/opentestsystem/rdw/common/model/ImportContent.java
```diff
@@ -10,7 +10,8 @@
     CODES(3),
     ORGANIZATION(4),
     GROUPS(5),
-    EMBARGO(6);
+    EMBARGO(6),
+    PERCENTILE(7);
```
malaffoon: comment from other review ... can we make this `NORMS(7)` or `NORMATIVE_DATA(7)`?

model/src/main/java/org/opentestsystem/rdw/common/model/ImportContent.java
```diff
@@ -10,7 +10,8 @@
     CODES(3),
     ORGANIZATION(4),
     GROUPS(5),
-    EMBARGO(6);
+    EMBARGO(6),
+    NORMS(7);
```
chdietz: change to NORMS

PR #60, opened 2018-02-02
Add codes to assessment type and subject enums.

wtritch:
While converting ReportQuery properties from DB ids to codes I find a few places in the code and many in the tests where we are using the AssessmentType and Subject enums to test/populate values.  I'd like to be able to test/populate codes rather than use static strings like "sum."

### DIFF
--- a/model/src/main/java/org/opentestsystem/rdw/common/model/AssessmentType.java
+++ b/model/src/main/java/org/opentestsystem/rdw/common/model/AssessmentType.java
@@ -9,15 +9,17 @@ package org.opentestsystem.rdw.common.model;
  * </p>
  */
 public enum AssessmentType {
-    ICA(1, true),
-    IAB(2, false),
-    SUMMATIVE(3, true);
+    ICA(1, "ica", true),
+    IAB(2, "iab", false),
+    SUMMATIVE(3, "sum", true);
 
     private final int id;
+    private final String code;
     private final boolean hasClaims;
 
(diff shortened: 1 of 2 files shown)


agorina:
I understand the convenience, but I am concern that we are duplicating codes in Java, especially for the Subjects (due to 'configurable subject' feature).

wtritch:
I agree, I'd love it this was only used in testing.  But we have places in the code where we do use the AssessmentType enum for logic already.  I'm not looking to introduce any more, but I do want to be able to convert the places we already have it to be able to switch on codes.  Once we have configurable subjects I fully expect the Subject enum to go away, but that's going to be a much heavier lift than I'd like to roll into my next PR.

wtritch:
For example, I want to change:
```
checkArgument(query.getAssessmentTypeId() != AssessmentType.IAB.id(), "custom aggregate report for IAB is not supported yet");
```
into
```
checkArgument(!AssessmentType.IAB.code().equals(query.getAssessmentTypeCode()), "custom aggregate report for IAB is not supported yet");
```
rather than
```
checkArgument(!"iab".equals(query.getAssessmentTypeCode()), "custom aggregate report for IAB is not supported yet");
```

agorina:
Discussed offline - approving as a temporary measure to get us moving.

esotrope:
And eventually it could be:
```java
checkArgument(!assessmentTypeService.getAssessmentType(query.getAssessmentTypeCode()).isInterim(), "ERROR !");
```

PR #61, opened 2018-02-11
DWR-1406 Added caseInsensitiveValue to Subject.

agorina:


agorina:
Merging, but will address comments if needed.

model/src/test/java/org/opentestsystem/rdw/common/model/SubjectTest.java
```diff
@@ -29,11 +35,6 @@ public void itShouldConvertFromString() {
         assertThat(Subject.valueOf("ELA")).isEqualTo(Subject.ELA);
     }
 
-    @Test(expected = IllegalArgumentException.class)
-    public void stringValueIsCaseSensitive() {
-        Subject.valueOf("ela");
-    }
-
```
malaffoon: why did you remove this test? Doesn't it demonstrate that valueOf() is case sensitive?

agorina: Oops, I do not know :)  I will restore it back.

PR #64, opened 2018-04-27
Add SUBJECT ImportContent type.

wtritch:


model/src/main/java/org/opentestsystem/rdw/common/model/ImportContent.java
```diff
@@ -11,7 +11,9 @@
     ORGANIZATION(4),
     GROUPS(5),
     EMBARGO(6),
-    NORMS(7);
+    NORMS(7),
+    SUBJECT(8),
+    LABEL(9);
```
wtritch: I added LABEL as a root import type.  The idea is that I want to be able to:
1. import a label (e.g. via a subject upload)
2. Migrate that label to reporting as a translation key (Math.key=value1, updated=8:00am)
3. Update the label in the reporting translation table (Math.key=value2, updated=8:30am)

Then we can avoid overwriting on a subject update via:
1. re-import the subject without changing`Math.key` from value1
2. detect that the label `Math.key` hasn't changed, do not update
3. on next migration do not migrate Math.key=value1 to reporting
4. Reporting still has Math.key=value2

OR overwrite via subject upload
1. re-import the subject after changing`Math.key` to value3
2. detect that the label `Math.key` has changed, update it
3. on next migration migrate Math.key=value3 to reporting
4. Reporting now has Math.key=value3

Thoughts?

agorina: Could you please clarify why would we need to update the label in the reporting translations table (bypassing the warehouse update)?

wtritch: After discussion I think we're going to suggest that modifying the reporting translations DB directly is not the recommended way to modify translations.  The recommended way to modify translations is to include an "en.json" (or xx.json for different languages) external translation file.
We are also going to move away from a generic "labels" concept in warehouse in favor of "subject_labels" bound to a subject for now.  If we introduce an additional feature with labels into warehouse in the future we can revisit a more generic "labels" warehouse concept.

PR #65, opened 2018-05-07
add group module for CSV validation

malaffoon:
Although i may have to tweak things i think it is pretty close. And i realized i have code modification across all four projects and that isn't good.


group/src/main/java/org/opentestsystem/rdw/group/CsvValidationService.java
```diff
@@ -0,0 +1,44 @@
...
+     * Validate a student group and return the collection of validation results. These may be
+     * failures and successes, be sure to check. Also populate file properties including
+     * "newline" and "delimiter".
+     *
+     * @param is student group content
+     * @param permissionScope permission scope for user validating content
+     * @param fileProperties container for file properties
+     * @return validation results
+     */
+    List<CsvValidationResult> validate(InputStream is, PermissionScope permissionScope, Map<String, String> fileProperties);
```
malaffoon: I changed the interface: before it accepted a CsvRecord iterator and there were separate methods for the headers and records. I wanted it to be opaque, just taking a data stream and returning results. The fileProperties thing is ugly but i didn't see an obvious way to deal with it without streaming the contents more than once.

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,333 @@
...
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static org.apache.commons.text.StringEscapeUtils.escapeJava;
+import static org.opentestsystem.rdw.group.CsvValidationResult.failure;
+
+/**
+ * Default implementation of a CsvValidationService
+ */
+@Service
+public class DefaultCsvValidationService implements CsvValidationService {
```
malaffoon: surprisingly little changed in here

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,333 @@
...
+    private static final String MoreFailuresFormat = "With %1$d additional validation failures";
+    private static final String MoreSuccessesFormat = "With %1$d additional schools";
+    private static final int MaxMessages = 100;
+    private final SchoolRepository schoolRepository;
+    private final SchoolYearRepository schoolYearRepository;
+
+    @Autowired
+    public DefaultCsvValidationService(
+            @Qualifier("warehouseJdbcSchoolRepository") final SchoolRepository schoolRepository,
+            @Qualifier("warehouseJdbcSchoolYearRepository") final SchoolYearRepository schoolYearRepository) {
```
malaffoon: note the qualifiers, more on this later

group/src/main/java/org/opentestsystem/rdw/group/JdbcSchoolRepository.java
```diff
@@ -0,0 +1,48 @@
...
+@Repository("warehouseJdbcSchoolRepository")
+class JdbcSchoolRepository implements SchoolRepository {
+
+    private final NamedParameterJdbcTemplate template;
+
+    @Value("${sql.school.findAll}")
+    private String findAllQuery;
+
+    @Autowired
+    JdbcSchoolRepository(@Qualifier("warehouseJdbcTemplate") final NamedParameterJdbcTemplate template) {
```
malaffoon: The client that includes this needs to provide this specifically-named template.

group/src/main/java/org/opentestsystem/rdw/group/JdbcSchoolYearRepository.java
```diff
@@ -0,0 +1,29 @@
...
+@Repository("warehouseJdbcSchoolYearRepository")
+class JdbcSchoolYearRepository implements SchoolYearRepository {
+
+    private final NamedParameterJdbcTemplate template;
+
+    @Value("${sql.schoolYear.findAll}")
+    private String findAllQuery;
+
+    @Autowired
+    JdbcSchoolYearRepository(@Qualifier("warehouseJdbcTemplate") final NamedParameterJdbcTemplate template) {
```
malaffoon: ditto

group/src/test/java/org/opentestsystem/rdw/group/DefaultCsvValidationServiceTest.java
```diff
@@ -0,0 +1,389 @@
...
+
+    private static class RecordBuilder {
+        final List<List<String>> records = newArrayList();
+
+        private static final CSVFormat format = CSVFormat.DEFAULT
+                .withIgnoreHeaderCase()
+                .withFirstRecordAsHeader()
+                .withHeader(RequiredHeaders.toArray(new String[0]));
+
+        String build() {
```
malaffoon: slight change in here if you're keeping close tabs

group/src/main/java/org/opentestsystem/rdw/group/CsvSubject.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.group;
+
+/**
+ * This class represents the Group subjects allowed during CSV Group uploads.
+ */
+public enum CsvSubject {
```
agorina: I know you just moved it, and I admit that this comment should have been made when this class was introduced, but since you are changing it: why do we need `CsvSubject` enum when we have `Subject` enum?  Could they be different?

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,333 @@
...
+            joiner.add(String.format(ValidationFailureFormat, validationResult.getRow(), validationResult.getMessage()));
+        }
+        if (failures.size() > MaxMessages) {
+            joiner.add(String.format(MoreFailuresFormat, failures.size() - MaxMessages));
+        }
+        return joiner.toString();
+    }
+
+    @Override
+    public String toSuccessMessage(final List<CsvValidationResult> successes) {
```
agorina: How is this different from the above method?  Would just one ' toMessage' be enough?

group/src/main/java/org/opentestsystem/rdw/group/CsvSubject.java
malaffoon: CsvSubject allows for `All`. I could change things to use the common `Subject` enum and then add in the special `All` handling separately but that would be awkward. I think the javadocs are pretty clear that this is for dealing with what is allowed in the CSV.

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
malaffoon: Yeah, i headed that way too but the text and data in the messages are slightly different.

group/src/main/java/org/opentestsystem/rdw/group/CsvSubject.java
wtritch: shoot... we're going to have to pull the list of subjects from the DB rather than rely on this enum

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,333 @@
...
+            validationResults.add(failure(0, "Error reading file [" + exception.getMessage() + "]"));
+        } catch (final IllegalArgumentException exception) {
+            logger.warn("Problem parsing group file", exception);
+            validationResults.add(failure(0, "File does not appear to be a valid CSV with a header row"));
+        }
+
+        return validationResults;
+    }
+
+    List<CsvValidationResult> validateHeaders(final List<String> csvHeaders) {
```
wtritch: Should these be private?  Or are they directly-accessed in a unit test?

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,333 @@
...
+        if (failures.isEmpty()) {
+            successes.add(success(recordNumber, groupCountPerSchool, currentSchool));
+            return ImmutableList.copyOf(successes);
+        }
+
+        return ImmutableList.copyOf(failures);
+    }
+
+    @Override
+    public String toFailureMessage(final List<CsvValidationResult> failures) {
```
wtritch: nit: public overrides should be at the top =)

group/src/main/java/org/opentestsystem/rdw/group/JdbcSchoolRepository.java
wtritch: ...Or instantiate it manually as a `@Bean` in a `@Configuration` class =)

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
malaffoon: They are directly accessed by a unit test. Basically, i left all the existing tests which exercised these helper methods.

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
malaffoon: Good point, i'll move 'em.

group/src/main/java/org/opentestsystem/rdw/group/CsvSubject.java
malaffoon: image

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,348 @@
...
+    private final Collection<Integer> schoolYears;
+    private final Collection<String> subjectCodes;
+
+    @Autowired
+    public DefaultCsvValidationService(
+            @Qualifier("warehouseJdbcSchoolRepository") final SchoolRepository schoolRepository,
+            @Qualifier("warehouseJdbcSchoolYearRepository") final SchoolYearRepository schoolYearRepository,
+            @Qualifier("warehouseJdbcSubjectRepository") final SubjectRepository subjectRepository) {
+        this.schoolRepository = schoolRepository;
+        this.schoolYears = schoolYearRepository.findAll();
```
malaffoon: @wtritch, @agorina can you take a quick look again? I changed it to use a repo to lookup subject codes and got rid of `CsvSubject`.

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
agorina: I am still confused: the current implementations of there two methods look identical (unless I am missing something). Why not have a helper method?

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
agorina: Looks good to me.

group/src/main/java/org/opentestsystem/rdw/group/DefaultCsvValidationService.java
```diff
@@ -0,0 +1,347 @@
...
+            if (failure.getMessage().equals(message)) return;
+        }
+        failures.add(failure(recordNumber, message));
+    }
+
+    private static CsvValidationResult success(final long recordNumber, final int groupCount, final String schoolId) {
+        return CsvValidationResult.success(recordNumber,
+                String.format(ValidationSuccessFormat, groupCount, groupCount == 1 ? "" : "s", schoolId));
+    }
+
```
malaffoon: @agorina i factored out a message helper. It kind of demonstrates what bothers me: note line 340 where we are formatting a message string for successes even though it doesn't need it (see the call above that passes "%2$s" as the message format). It does reduce code i guess.

PR #66, opened 2018-05-08
Add subject definition model and de/serialization utility

wtritch:
This PR adds the Subject definition model and a de/serialization utility that validates the XML payload against the XSD while streaming the source to Jackson for de-serialization as a model.


build.gradle
```diff
@@ -51,6 +51,12 @@ subprojects {
       dependency 'junit:junit:4.12'
       dependency 'org.assertj:assertj-core:3.6.2'
       dependency 'org.mockito:mockito-core:2.16.0'
+
+      // Woodstox STAX2 XML stream processing API and required
+      // Sun Multi-Schema Validation libraries
+      dependency 'org.codehaus.woodstox:stax2-api:4.1'
+      dependency 'msv:msv:20050913'
+      dependency 'msv:xsdlib:20050913'
```
wtritch: Jackson and XSD validation don't play well together.  Or at all. After some googling I found that woodstox' stax2 implementation was one of the only streaming, validating xml libraries around.

model/src/main/java/org/opentestsystem/rdw/common/model/subject/CodedEntity.java
```diff
@@ -0,0 +1,38 @@
+package org.opentestsystem.rdw.common.model.subject;
+
+/**
+ * Base class representing a Subject entity that contains a code property.
+ */
+public abstract class CodedEntity {
```
wtritch: I javadoc'd the Subject definition models.

model/src/main/java/org/opentestsystem/rdw/common/model/subject/SubjectXmlUtils.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param inputStream   An XML input stream
+     * @return  The Subject represented by the given XML
+     */
+    public static Subject subjectFromXml(final InputStream inputStream) {
+        try {
+            return getMapper().readValue(getStreamReader(inputStream), Subject.class);
+        } catch (final IOException e) {
+            throw new RuntimeException(e);
+        }
```
wtritch: This is the interesting bit: given an InputStream containing the XML bytes, I wrap it in a validating XMLStreamReader that validates against the XSD as it streams the XML to Jackson for deserialization.
The benefit: I don't need to double-check that we're using valid color names, assessment type codes, grade codes, etc that are all validated by the XSD.

model/src/main/java/org/opentestsystem/rdw/common/model/subject/SubjectMessage.java
```diff
@@ -0,0 +1,50 @@
+package org.opentestsystem.rdw.common.model.subject;
+
+/**
+ * This class represents a semi-structured key/value message pair for the parent
+ * subject.
+ */
+public class SubjectMessage {
```
esotrope: so i'm curious - what does the team feel about xsd pojo generation libraries?

wtritch: I believe @malaffoon did that with the TRT models.  Buuuut since I already had the mixins set, I dislike annotating models with serialization information, and we're free to switch to JSON with a separate set of mixins... I prefer this approach, although I'm not going to go refactor the TRT handling.

model/src/main/java/org/opentestsystem/rdw/common/model/subject/ItemClaim.java
```diff
@@ -0,0 +1,145 @@
...
+
+import java.util.List;
+
+/**
+ * This class represents "Claim" and/or "Scorable Claim" in the reporting system
+ * for the parent Subject.
+ * A claim may be scorable, organizational, or both.
+ * A claim cannot be neither scorable nor organizational.
+ */
+public class ItemClaim extends NamedCodedEntity {
```
malaffoon: `Claim` too generic?

model/src/main/java/org/opentestsystem/rdw/common/model/subject/ItemDifficulty.java
```diff
@@ -0,0 +1,71 @@
+package org.opentestsystem.rdw.common.model.subject;
+
+/**
+ * This class represents assessment item difficulty cutpoints
+ * within the context of a grade.
+ * An item's difficulty is calculated via comparing it's numeric difficulty value
```
malaffoon: s/it's/its   (or, as i like to call it, my Will macro)

wtritch: /sigh fixed

model/src/main/java/org/opentestsystem/rdw/common/model/subject/ItemClaim.java
wtritch: It felt a little generic... but I can change it if you like?

model/src/main/java/org/opentestsystem/rdw/common/model/subject/jackson/CodedEntityMixin.java
```diff
@@ -0,0 +1,12 @@
+package org.opentestsystem.rdw.common.model.subject.jackson;
+
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
+
+/**
+ *
```
malaffoon: It would be nice to have `@see` comments that link the mixin to the matching class

model/src/main/java/org/opentestsystem/rdw/common/model/subject/SubjectXmlUtils.java
malaffoon: thanks for figuring this out

model/src/main/java/org/opentestsystem/rdw/common/model/subject/jackson/CodedEntityMixin.java
wtritch: Good call, done

model/src/main/java/org/opentestsystem/rdw/common/model/subject/SubjectXmlUtils.java
wtritch: np, only felt like throwing my computer out of a window a couple of times =)

model/build.gradle
```diff
@@ -1,2 +1,13 @@
...
+dependencies {
+    compile 'com.fasterxml.jackson.core:jackson-annotations'
+    compile 'com.fasterxml.jackson.core:jackson-databind'
+    compile 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml'
+    compile 'com.fasterxml.jackson.datatype:jackson-datatype-guava'
+    compile 'org.codehaus.woodstox:stax2-api'
+    compile 'msv:msv'
+    compile 'msv:xsdlib'
+
+    runtime files(Jvm.current().toolsJar)
```
wtritch: I just ran into a missing class definition issue when trying to run the gradle build via command-line:
```
org.opentestsystem.rdw.common.model.subject.SubjectXmlUtilsTest > itShouldSerializeAndDeserializeSubjectAsXML FAILED
    java.lang.NoClassDefFoundError: org/relaxng/datatype/ValidationContext
        at java.lang.ClassLoader.defineClass1(Native Method)
```
For whatever reason, Gradle doesn't add the `tools.jar` to the classpath when running via command-line unless you tell it to.  The test ran fine in IntelliJ...

PR #68, opened 2018-05-10
move tests from reporting to common

malaffoon:
I found some tests in reporting that belong here. I'll be deleting them from reporting.


malaffoon:
Since these are just tests and they seem to work, i'm going to merge and address issues later.

PR #70, opened 2018-05-11
Add shared constant model values.

wtritch:
This PR adds shared constant values to the RDW_Common model module.


model/src/main/java/org/opentestsystem/rdw/common/model/ModelConstants.java
```diff
@@ -0,0 +1,20 @@
+package org.opentestsystem.rdw.common.model;
+
+/**
+ * This class holds system-defined model-scoped values that are accessible across
+ * the RDW ecosystem, but are not configurable.
+ */
+public abstract class ModelConstants {
```
wtritch: This feels a little ham-fisted, but I can't think of a better option...  Thoughts?

jtreuting: Maybe we make it specific for configurable subjects or assessment definitions at this point since that is the need.  Somehting like `AssessmentBoundaries` or `AssessmentMetaDataConstants`.  Not sure I love these but maybe making it a little more narrow is appropriate instead of a cover everything type class.

wtritch: Hm... except it's bound more to Subjects than to Assessments.  Maybe SubjectConstants?

wtritch: well... except it also affects the subject/assessment pairs... I hate naming.  If you want `AssessmentBoundaries`, that works for me.

agorina: I like the word 'Boundaries' in the name, as for the Assessment or System or anything else - I could see it either way.

PR #72, opened 2018-05-17
Add AssessmentType.fromCode to lookup an enum by the code

jtreuting:
For the student report processor to handle ICA and Summative we have the code from the request and i'd like to use this to get the AssessmentType enum and therefore the ID for the assessment type to use in the SQL.


jtreuting:
going to merge so i can reference it in the other code.  if there are strong opinions i'll address after

PR #77, opened 2018-06-15
make some of the TRT attributes optional

malaffoon:


model/src/main/resources/TestResultsTransmissionFormat.xsd
```diff
@@ -23,7 +23,7 @@ governing permissions and limitations under the License.
             <xs:attribute name="testId" use="required"/>
             <xs:attribute name="bankKey" type="xs:unsignedInt"/>
             <xs:attribute name="contract"/>
-            <xs:attribute name="mode" use="required">
+            <xs:attribute name="mode" default="online">
```
malaffoon: For the enum attributes, i chose to make a default for each.

model/src/test/java/org/opentestsystem/rdw/common/model/trt/XmlUtilsTest.java
```diff
@@ -66,6 +66,27 @@ public void itShouldUnmarshalSamplesFromTDS() {
...
+
+    @Test
+    public void itShouldUnmarshalMinimalSample() {
+        final TDSReport tdsReport = XmlUtils.tdsReportFromXml(this.getClass().getResourceAsStream("/TDSReport.minimal.xml"));
+        assertThat(tdsReport.getTest().getGrade()).isEqualTo("7");
+        // test attributes with default values
+        assertThat(tdsReport.getTest().getMode()).isEqualTo("online");
+        assertThat(tdsReport.getOpportunity().getStatus()).isEqualTo("scored");
+        assertThat(tdsReport.getOpportunity().getValidity()).isEqualTo("valid");
+    }
```
malaffoon: i might come back and assert all the null values but i figured that isn't necessary to move forward

model/src/main/resources/TestResultsTransmissionFormat.xsd
malaffoon: I will forward these changes to Alex once we vet the entire processing pipeline.

model/src/main/java/org/opentestsystem/rdw/common/model/trt/Opportunity.java
```diff
@@ -497,7 +498,7 @@ public void setCompleteness(String value) {
      *
      * @return opportunity
      */
-    public long getOpportunity() {
+    public Long getOpportunity() {
```
jtreuting: is this different than oppId which we said is required?

malaffoon:
I'm going to merge so we can test downstream stuff

PR #80, opened 2018-06-26
Update valid subject colors to allow for new iab colors.

wtritch:


model/src/main/resources/RDW_Subject.xsd
```diff
@@ -430,7 +430,9 @@
       <xs:enumeration value="gray-lighter"/>
       <xs:enumeration value="gray-lightest"/>
       <xs:enumeration value="white"/>
-      <xs:enumeration value="textured-blue"/>
```
wtritch: I'm removing textured-blue because it doesn't function properly as a label/text/etc color and we no longer require it as a fill color for IABs.

PR #81, opened 2018-06-27
Removed hardcoded values.

agorina:


model/src/main/java/org/opentestsystem/rdw/common/model/trt/Opportunity.java
```diff
@@ -796,7 +795,7 @@ public void setCompleteStatus(String value) {
      *
      */
     public String getAdministrationCondition() {
-        return (administrationCondition == null || administrationCondition.isEmpty()) ? "SD" : administrationCondition;
```
agorina: This does not work with the optional data elements.

agorina:
Merging since I need if for the ingest project. Will address comments if needed.

PR #82, opened 2018-06-28
Removed subject enum.

agorina:


agorina:
Merging since I doubt there will be any objections, but will address if there are.

PR #84, opened 2018-07-03
Add an optional longName display name to an assessment type within th…

wtritch:
…e context of a subject.


model/src/main/resources/RDW_Subject.xsd
```diff
@@ -90,13 +99,19 @@
...
     </xs:complexType>
   </xs:element>
   <xs:element name="PerformanceLevel">
     <xs:complexType>
-      <!-- The integer value of the level. -->
+      <!--
+        The integer value of the level.
+        NOTE: Performance levels must start at 1 and continue sequentially up to 6
+        NOTE: iab assessment types only support 3 performance levels
+      -->
```
wtritch: Moved over some docs that were in the RDW version but hadn't made it here yet.

model/src/main/resources/RDW_Subject.xsd
```diff
@@ -65,13 +65,22 @@
...
       </xs:sequence>
       <!-- The assessment type being defined. Must be "ica", "iab" or "sum" -->
       <xs:attribute name="code" use="required" type="AssessmentTypeCode"/>
       <!-- This is the display name for the assessment type. -->
       <xs:attribute name="name" use="required" type="xs:string"/>
+      <!--
+        This is an optional longer version of the display name for the assessment type.
+        If not provided, it will default to the "name" value.
+      -->
+      <xs:attribute name="longName" type="xs:string"/>
```
wtritch: This is the change allowing a "longName" attribute on an AssessmentType element

PR #87, opened 2018-09-09
Add helper to apply an XSL transformation when reading a TDSReport.

malaffoon:
This is a supporting change for an enhancement that allows a transformation to be applied to TRT files as they are processed. Please see https://github.com/SmarterApp/RDW_Ingest/pull/546 for more details.

I did a little informal benchmarking. With small transformations (like the example), reading and converting a TRT to the TDSReport data model goes from about 950ms to 1200ms. So it isn't something we want to do if we don't need to (you'll see how that is handled in the RDW_Ingest changes).

Sorry to drag you into these reviews but i need critical review since it is urgent and we don't have a lot of resources to apply.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.class
 build/
+out/
 
 *.iml
 .idea/
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ apply plugin: 'io.spring.dependency-management'
 
 allprojects {
     group = 'org.opentestsystem.rdw.common'
-    version = '1.2.0-SNAPSHOT'
+    version = '1.2.2-SNAPSHOT'
 
     apply plugin: 'idea'
     apply plugin: 'propdeps'
--- a/model/src/main/java/org/opentestsystem/rdw/common/model/trt/XmlUtils.java
+++ b/model/src/main/java/org/opentestsystem/rdw/common/model/trt/XmlUtils.java
@@ -1,5 +1,7 @@
 package org.opentestsystem.rdw.common.model.trt;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.xml.sax.SAXException;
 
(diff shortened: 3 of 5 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 *.class
 build/
+out/
```
malaffoon: unrelated, just bugging me

build.gradle
```diff
@@ -15,7 +15,7 @@ apply plugin: 'io.spring.dependency-management'
 
 allprojects {
     group = 'org.opentestsystem.rdw.common'
-    version = '1.2.0-SNAPSHOT'
+    version = '1.2.2-SNAPSHOT'
```
malaffoon: HOTFIX v1.2.2

model/src/test/resources/example.xsl
```diff
@@ -0,0 +1,34 @@
+<xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
```
malaffoon: Just some example transformations. I'll probably copy some of this to the main RDW documentation.

model/src/main/java/org/opentestsystem/rdw/common/model/trt/XmlUtils.java
```diff
@@ -1,33 +1,55 @@
...
+     *
+     * @param is XML stream
+     * @param xsl XSLT stream
+     * @return TDSReport
+     */
+    public static TDSReport tdsReportFromXml(final InputStream is, final InputStream xsl) {
+        try {
+            final JAXBResult result = new JAXBResult(createUnmarshaller());
+            final Transformer transformer = getTransformer(xsl);
+            transformer.transform(new StreamSource(is), result);
```
malaffoon: Since i stared at this for 10 minutes trying to figure it out, i'll just mention: `transform` accepts a JAXBResult and modifies it in place.

PR #89, opened 2018-12-20
Add targetReport to SubjectAssessmentType

malaffoon:
I'm going to use this as i implement the import processing so i may find a need to make changes to support that.


model/src/main/java/org/opentestsystem/rdw/common/model/subject/XmlSubjectSerializationService.java
```diff
@@ -27,7 +28,16 @@
...
-        return getMapper().readValue(getStreamReader(inputStream), Subject.class);
+        final Subject subject = getMapper().readValue(getStreamReader(inputStream), Subject.class);
+        // The desired functionality is an optional attribute with a default value that is an
+        // expression depending on the 'code' attribute. Since XSD doesn't support that, this
+        // post-processing provides equivalent functionality.
+        for (final SubjectAssessmentType subjectAssessmentType : subject.getAssessmentTypes()) {
+            if (subjectAssessmentType.isTargetReport() == null) {
+                subjectAssessmentType.setTargetReport(subjectAssessmentType.getCode().equals(AssessmentType.SUMMATIVE.code()));
+            }
+        }
```
malaffoon: I'm not thrilled with this approach but the alternatives are all messy (i played with transforming the XML on the way in, or fiddling with the node tree during processing).

burzum619: I'm good with this approach.

model/src/test/java/org/opentestsystem/rdw/common/model/subject/XmlSubjectSerializationServiceTest.java
```diff
@@ -125,11 +161,12 @@ private Target target(final String claimCode, final String targetCode) {
                 .build();
     }
 
-    private SubjectAssessmentType assessmentType(final String typeCode, final String name) {
+    private SubjectAssessmentType assessmentType(final String typeCode, final String name, Boolean targetReport) {
```
burzum619: nit: targetReport param should be `final`

malaffoon: thanks, fixed

PR #93, opened 2019-03-25
Force printedReport flag to default true

malaffoon:
Apparently the `default=` spec in an XSD isn't required to be honored. And stax2 doesn't. So we have to force the default value in the data model.


model/src/test/java/org/opentestsystem/rdw/common/model/subject/XmlSubjectSerializationServiceTest.java
```diff
@@ -157,6 +159,18 @@ public void itShouldLoadELPAC() {
         }
     }
 
+    @Test
+    public void itShouldDefaultPrintedReportTrue() {
+        // reusing another test file
+        try (final InputStream is = this.getClass().getResourceAsStream("/ClaimOrder.xml")) {
+            final Subject subject = service.parseSubject(is);
+            final SubjectAssessmentType asmtType = subject.getAssessmentTypes().get(0);
+            assertThat(asmtType.isPrintedReport()).isTrue();
```
malaffoon: This was failing before the fix.

PR #94, opened 2019-04-02
review feedback

malaffoon:


multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/AbstractDynamicRoutingDataSource.java
```diff
@@ -52,12 +52,7 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
      * @return DataSource target
      */
     protected DataSource determineTargetDataSource() {
-        Object lookupKey = determineCurrentLookupKey();
-        if (!resolvedDataSources.containsKey(lookupKey)) {
-            //avoid problem of 2 simultaneous new calls to new tenant
-            resolvedDataSources.putIfAbsent(lookupKey, dataSourceFactory(lookupKey));
-        }
-        return resolvedDataSources.get(lookupKey);
+        return resolvedDataSources.computeIfAbsent(determineCurrentLookupKey(), this::dataSourceFactory);
```
malaffoon: `computeIfAbsent` is atomic so i'm pretty sure this simplification is safe.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/DataSourceUrlParts.java
```diff
@@ -7,7 +7,7 @@
  *
  * @see <a href="https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-jdbc-url-format.html">https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-jdbc-url-format.html</a>
  */
-public class DataSourceUrlPartsProperties {
+public class DataSourceUrlParts {
```
malaffoon: Recommending this for two reasons: 
1. There is a member in here that is called properties and i just think code like `properties.getProperties()` will be weird.
2. In Spring XxxProperties classes are generally containers for configuration properties, i.e. property beans. Since i don't quite know how this will be used that is either an argument for or against the renaming. If it is being used a property bean, then i should revert this.

multi-tenant/src/test/java/org/opentestsystem/rdw/multitenant/AbstractDynamicRoutingDataSourceTest.java
```diff
@@ -4,14 +4,18 @@
 
 import javax.sql.DataSource;
 
-import static org.mockito.Mockito.*;
```
malaffoon: Convention for this codebase is to not use * in imports. Suggest tweaking your IJ setting to change import counts to 99.

multi-tenant/src/test/java/org/opentestsystem/rdw/multitenant/AbstractDynamicRoutingDataSourceTest.java
```diff
@@ -4,14 +4,18 @@
...
 
 public class AbstractDynamicRoutingDataSourceTest {
 
-    DataSource dataSourceA = mock(DataSource.class);
-    DataSource dataSourceB = mock(DataSource.class);
+    private final DataSource dataSourceA = mock(DataSource.class);
+    private final DataSource dataSourceB = mock(DataSource.class);
 
-    AbstractDynamicRoutingDataSource dynamicRoutingDataSource = new AbstractDynamicRoutingDataSource() {
+    private AbstractDynamicRoutingDataSource dynamicRoutingDataSource = new AbstractDynamicRoutingDataSource() {
```
malaffoon: Make IJ happy.

multi-tenant/src/test/java/org/opentestsystem/rdw/multitenant/AbstractDynamicRoutingDataSourceTest.java
```diff
@@ -45,12 +49,18 @@ public void testDynamicDatasourceA() throws Exception {
     @Test
     public void testDynamicDatasourceB() throws Exception {
         TenantContextHolder.setTenantId("B");
-        dynamicRoutingDataSource.getConnection();
-        verify(dataSourceB).getConnection();
-        verify(dataSourceA, times(0)).getConnection();
+        dynamicRoutingDataSource.getConnection("username", "password");
```
malaffoon: This change is really for coverage.

multi-tenant/src/test/java/org/opentestsystem/rdw/multitenant/TenantDatasourceOverrideTest.java
```diff
@@ -0,0 +1,16 @@
+package org.opentestsystem.rdw.multitenant;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TenantDatasourceOverrideTest {
+
+    @Test
```
malaffoon: Adding coverage; i know it's a trivially simple class but i have been burned in the past by the order of the arguments being coded wrong.

multi-tenant/src/test/java/org/opentestsystem/rdw/multitenant/TenantDynamicRoutingDataSourceTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class TenantDynamicRoutingDataSourceTest {
+
+    private DataSourceBuilder dataSourceBuilder;
+    private DataSourceProperties dataSourceProperties;
+    private TenantDataSourceOverrideResolver tenantDataSourceOverrideResolver;
+    private TenantDynamicRoutingDataSource tenantDataSource;
+
+    @Before
+    public void createDataSource() {
```
malaffoon: Arguably if the tests require this much set up, then the tests might be too knowledgable of the implementation. I could be talked out of this test suite. But RDW_Common has pretty good coverage so i added it to keep the trend going.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/DataSourceUrlParts.java
ftdholbrook: Originally was going to drive from from spring config server.  Still on the bubble, but I'm fine with the simpler name regardless.

multi-tenant/src/test/java/org/opentestsystem/rdw/multitenant/AbstractDynamicRoutingDataSourceTest.java
ftdholbrook: ok

PR #98, opened 2019-04-30
RP-277	Common - Make S3 Interactions multi-tenant aware

vgonzalez9:
Add tenant path info to S3 and local archives.
Default tenant path is ""
TODO hook in with real tenant Info - Up for Dave for initial discussions

### DIFF
--- a/archive/build.gradle
+++ b/archive/build.gradle
@@ -2,6 +2,7 @@ dependencies {
     compile 'com.google.guava:guava'
     compile 'org.springframework:spring-core'
     compile 'org.springframework.cloud:spring-cloud-starter-aws'
+    compile project(':rdw-common-multi-tenant')
 
     testCompile 'org.springframework:spring-test'
     testRuntime 'org.springframework.boot:spring-boot-starter-test'
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveConfiguration.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveConfiguration.java
@@ -4,6 +4,7 @@ import com.amazonaws.auth.BasicAWSCredentials;
 import com.amazonaws.internal.StaticCredentialsProvider;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
+import org.opentestsystem.rdw.multitenant.storage.TenantPathDefaultResolver;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.boot.bind.RelaxedPropertyResolver;
 import org.springframework.context.annotation.Bean;
@@ -60,6 +61,9 @@ public class ArchiveConfiguration {
             final String pre = prefix == null ? "" : prefix.endsWith(".") ? prefix : (prefix + ".");
             PropertyResolver resolver = new RelaxedPropertyResolver(propertyResolver, pre);
             final String root = resolver.getRequiredProperty("root");
+            // Add tenant awareness
+            TenantPathDefaultResolver tenantPathNameResolver = new TenantPathDefaultResolver();
+
             if (LocalArchiveService.validRoot(root)) {
                 return localArchiveService(root);
             }
@@ -73,7 +77,9 @@ public class ArchiveConfiguration {
                 resolver = new RelaxedPropertyResolver(propertyResolver, pre + "cloud.aws.region.");
                 builder.setRegion(resolver.getRequiredProperty("static"));
                 final String sse = new RelaxedPropertyResolver(propertyResolver, pre + "cloud.aws.").getProperty("sse");
-                return s3ArchiveService(builder.build(), root, sse);
+
+
+                return s3ArchiveService(builder.build(), root, tenantPathNameResolver.getPrefix(), sse);
             }
             throw new IllegalArgumentException("Invalid archive root " + root);
         };
@@ -81,8 +87,8 @@ public class ArchiveConfiguration {
 
     @Bean
     @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
-    public S3ArchiveService s3ArchiveService(final AmazonS3 amazonS3, final String archiveRoot, final String sse) {
-        return new S3ArchiveService(amazonS3, archiveRoot, sse);
+    public S3ArchiveService s3ArchiveService(final AmazonS3 amazonS3, final String archiveRoot, final String prefix, final String sse) {
+        return new S3ArchiveService(amazonS3, archiveRoot, prefix, sse);
     }
 
     @Bean
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
@@ -34,6 +34,8 @@ import java.util.Properties;
  * @see ArchiveConfiguration
  */
 public class LocalArchiveService implements ArchiveService {
+
+
     private static final Logger logger = LoggerFactory.getLogger(LocalArchiveService.class);
     private static final String FileSchemePrefix = "file://";
 
@@ -44,6 +46,7 @@ public class LocalArchiveService implements ArchiveService {
             .build();
 
     /**
+     * expecting tenant to be handled in the passed in root
      * @param root root path, must start with "file://", e.g. "file:///tmp/"
      */
     public LocalArchiveService(final String root) {
@@ -54,9 +57,18 @@ public class LocalArchiveService implements ArchiveService {
         rootPath = path.endsWith(File.separator) ? path : (path + File.separator);
     }
 
+    /* tenant prefic in root
+    public String getTenantPath() {
+        // todo call TenantPathResolver
+        String tenantId = ""; // todo getFromTenantContext();
+        if (tenantId.isEmpty()) return "";
+        // todo add look up for Id
+        return "DEFAULT_TENANT/";
+    } */
+
     @Override
     public String getRawUri(final String location) {
-        return FileSchemePrefix + realResourceFile(location).getAbsolutePath();
+        return FileSchemePrefix +  realResourceFile(location).getAbsolutePath();
     }
 
     @Override
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
@@ -10,6 +10,7 @@ import com.amazonaws.services.s3.model.ObjectMetadata;
 import com.amazonaws.services.s3.model.S3ObjectSummary;
 import com.amazonaws.util.IOUtils;
 import com.google.common.base.CharMatcher;
+import org.opentestsystem.rdw.multitenant.storage.TenantPathDefaultResolver;
 import org.slf4j.Logger;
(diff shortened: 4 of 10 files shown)


archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -38,14 +40,16 @@
 
     private final AmazonS3 amazonS3;
     private final String bucket;
+    private final String objPrefix;
     private final String sse;
 
     /**
      * @param amazonS3 amazon S3 client
      * @param root root path, must start with "s3://", e.g. "s3://myBucket"
+     * @param objPrefix root path, must start with "s3://", e.g. "s3://myBucket"
```
malaffoon: javadocs are wrong

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -38,14 +40,16 @@
...
      * @param sse optional SSE algorithm, may be null, e.g. "AES256"
      */
-    public S3ArchiveService(final AmazonS3 amazonS3, final String root, final String sse) {
+    public S3ArchiveService(final AmazonS3 amazonS3, final String root, final String objPrefix, final String sse) {
```
malaffoon: It seems like we could avoid adding another argument. Instead, the caller could just extend the root. That is call `new S3ArchiveService(amazonS3, "s3://rdw-bucket/tenant_nv/", sse)` instead of `new S3ArchiveService(amazonS3, "s3://rdw-bucket", "tenant_nv", sse)`.

Still thinking it through because that does put a little more burden on this class to know about buckets and stuff. I think it would be cleaner though.

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -58,11 +62,14 @@ public S3ArchiveService(final AmazonS3 amazonS3, final String root, final String
...
         this.sse = sse;
+        this.objPrefix = objPrefix;
     }
 
+
     @Override
     public String getRawUri(final String location) {
-        return S3SchemePrefix + bucket + "/" + location;
+        // "" empty prefix is the default
+        return S3SchemePrefix + bucket + "/" + (objPrefix.isEmpty() ? "" : objPrefix + "/") + location;
```
malaffoon: This is assuming the caller knows what they are doing and did (or didn't) include a slash, etc. It seems like it would be better to do this logic of building up the proper prefix once in the constructor instead of for every call.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/storage/TenantPathResolver.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Strategy pattern to find the current tenant Path.  The goal being to abstract implementation
+ * variations away such as SecurityContext or TenantContextHolder
+ */
+public interface TenantPathResolver {
+
+    /**
+     * Provides a specific {@link TenantPathOverride} for a given tenant.
+     * expecting these to
```
malaffoon: now i really want to know what we're expecting  ;)

multi-tenant/src/test/java/org/opentestsystem/rdw/multitenant/storage/TenantPathNameDefaultResolverTest.java
```diff
@@ -0,0 +1,7 @@
+package org.opentestsystem.rdw.multitenant.storage;
```
malaffoon: going to complete these tests?

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveConfiguration.java
```diff
@@ -73,16 +77,18 @@ public ArchiveServiceFactory archiveServiceFactory() {
...
 
     @Bean
     @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
-    public S3ArchiveService s3ArchiveService(final AmazonS3 amazonS3, final String archiveRoot, final String sse) {
-        return new S3ArchiveService(amazonS3, archiveRoot, sse);
+    public S3ArchiveService s3ArchiveService(final AmazonS3 amazonS3, final String archiveRoot, final String prefix, final String sse) {
+        return new S3ArchiveService(amazonS3, archiveRoot, prefix, sse);
     }
 
     @Bean
```
malaffoon: It may be that maintaining and testing the LocalArchiveService is becoming more PITA than useful developer utility, but i think we have to maintain it as long as it exists. So the changes to S3 need to be reflected in the local archive service too.

archive/build.gradle
```diff
@@ -2,6 +2,7 @@ dependencies {
     compile 'com.google.guava:guava'
     compile 'org.springframework:spring-core'
     compile 'org.springframework.cloud:spring-cloud-starter-aws'
+    compile project(':rdw-common-multi-tenant')
```
malaffoon: Hmm, it feels like we might be coupling these two modules too much. Perhaps the `ArchiveConfiguration` needs to be tweaked so that it can pass through the tenant without being aware of tenant resolution. Or, maybe the whole ArchiveConfiguration thing needs to be pulled out of here and into the application context.

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveConfiguration.java
```diff
@@ -60,6 +61,9 @@ public ArchiveServiceFactory archiveServiceFactory() {
             final String pre = prefix == null ? "" : prefix.endsWith(".") ? prefix : (prefix + ".");
             PropertyResolver resolver = new RelaxedPropertyResolver(propertyResolver, pre);
             final String root = resolver.getRequiredProperty("root");
+            // Add tenant awareness
+            TenantPathDefaultResolver tenantPathNameResolver = new TenantPathDefaultResolver();
```
malaffoon: see comment above ... it feels wrong that this module is aware of tenant resolution.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/storage/TenantPathOverride.java
```diff
@@ -0,0 +1,19 @@
+package org.opentestsystem.rdw.multitenant.storage;
+
+public class TenantPathOverride {
+    private final String root;
+    private final String prefix;
```
malaffoon: Can you give an example of the root and prefix? I ask because i'm wondering if we can have a single configured value and just a convention for generating the other one.

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -54,9 +57,18 @@ public LocalArchiveService(final String root) {
         rootPath = path.endsWith(File.separator) ? path : (path + File.separator);
     }
 
+    /* tenant prefic in root
```
malaffoon: i'm assuming this is going away?

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -54,9 +57,18 @@ public LocalArchiveService(final String root) {
...
+        String tenantId = ""; // todo getFromTenantContext();
+        if (tenantId.isEmpty()) return "";
+        // todo add look up for Id
+        return "DEFAULT_TENANT/";
+    } */
+
     @Override
     public String getRawUri(final String location) {
-        return FileSchemePrefix + realResourceFile(location).getAbsolutePath();
+        return FileSchemePrefix +  realResourceFile(location).getAbsolutePath();
```
malaffoon: we could probably use that space somewhere else  ;)

archive/build.gradle
ftdholbrook: I think the only way clean to do it is to make the configuration external to the service.  That would take a big-ish refactor, but may be worth it.

malaffoon: I admit i'm still trying to get my head around how this is setup, so my comments may be misguided.

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveConfiguration.java
```diff
@@ -73,16 +77,18 @@ public ArchiveServiceFactory archiveServiceFactory() {
...
         };
     }
 
     @Bean
     @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
-    public S3ArchiveService s3ArchiveService(final AmazonS3 amazonS3, final String archiveRoot, final String sse) {
-        return new S3ArchiveService(amazonS3, archiveRoot, sse);
+    public S3ArchiveService s3ArchiveService(final AmazonS3 amazonS3, final String archiveRoot, final String prefix, final String sse) {
+        return new S3ArchiveService(amazonS3, archiveRoot, prefix, sse);
     }
```
ftdholbrook: Just to take a step back from the changes just made, I think this 
```
@Bean
public ArchiveServiceFactory archiveServiceFactory() {
```
Is doing far too much here.  Unless I'm missing something S3ArchiveServies the only reason instances here are prototype scoped is so multiple down steam apps can have their own versions that are otherwise identical?.  I think it may be cleaner overall to remove this configuration, make a clean bean based config object and have the consumers configure singletons as needed.

archive/build.gradle
malaffoon: Couldn't we just change the signature of `ArchiveServiceFactory#createArchiveService` to take the tenant prefix (or root). Then wrap ArchiveConfiguration with something that is tenant aware?

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveConfiguration.java
malaffoon: 👍

archive/build.gradle
ftdholbrook: It would have to be a strategy pattern and not just a string.

```
    @Bean
    public ArchiveService archiveService(final ArchiveServiceFactory factory, final Environment environment) {
        return factory.createArchiveService(environment, "archive.");
    }
```
Adding a string to something like that would not be tenant aware, and it it is a strategy, it will have to know about tenancy

alternatively something like this

```
   @Bean
   @Configureation("path.to.config")
    ArchiveServiveseElements 

    @Bean
    public ArchiveService archiveService(ArchiveServiceConfig elements) {
        return ArchiveService(elements)
    }
```
get rid of the whole factory thing and just use standard-ish spring stuff much like we are doing for datasources.  We could put the tenancy in the elements part.  make it an interface on the archive side and have a concrete tenant aware version on the multi-tenant side.

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
vgonzalez9: I agree it would be cleaner. Less overall changes. I went back and forth on this a few times, this obvious called out the tenant impact with the bucket. And I am good changing it to 
s3://{bucket}/{tenant}/{location} and pull out each piece as needed.

vgonzalez9:
Closing this pull request. Changing how the multi-tenancy is configured, and use updated configuration in a similar fashion to Datasource feature changes.

PR #102, opened 2019-05-11
RP-237: migrate script common from RDW_Ingest

gregcharles:
Not sure how they calculate the 27 commits. I guess it's the total number of commits made to script-common in RDW-Ingest, plus the few it took to get it migrated and working here.


script/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compile 'commons-io:commons-io'
+    compile 'jaxen:jaxen:1.1.4'
+    compile 'net.sf.saxon:Saxon-HE:9.7.0-21'
+    compile 'org.springframework.boot:spring-boot-starter-jdbc'
```
malaffoon: This needed?

script/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+
+    // for @NotNull
+    compile 'javax.validation:validation-api'
+
+    compile project(':rdw-common-utils')
+    compile project(':rdw-common-model')
+
+    testCompile 'org.xmlunit:xmlunit-core:2.6.2'
+    testCompile 'org.xmlunit:xmlunit-matchers:2.6.2'
+    testCompile 'org.springframework.boot:spring-boot-starter-test'
```
malaffoon: This needed?

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import org.opentestsystem.rdw.common.model.ImportException;
+import org.opentestsystem.rdw.script.security.SandboxSecurityManager;
+import org.opentestsystem.rdw.script.util.ScriptUtils;
+
+/**
+ * A Pipeline is a series of precompiled scripts that run against input in order to
+ * transform, filter, and validated it.
```
malaffoon: nit: should be `validate it.`

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
```diff
@@ -0,0 +1,138 @@
...
+
+        return new PipelineResults(lastValidTransformation);
+    }
+
+    private Object runScript(final Script script, final ResourceType type) {
+        if (type == ResourceType.SYSTEM_NODE || sandboxSecurityManager == null) {
+            return script.run();
+        }
+
+        return runInSandbox(script, SANDBOX_RETRIES);
```
malaffoon: nit: i don't really see the advantage of lines 85-88; just "inline" it?

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    private Object runInSandbox(final Script script, final int retries) {
+        final Callable<Object> callable = () -> {
```
malaffoon: Could just inline this in the `submit` call; i think it would be more readable that way.

script/src/main/java/org/opentestsystem/rdw/script/PipelineBuildException.java
```diff
@@ -0,0 +1,28 @@
+package org.opentestsystem.rdw.script;
+
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * Exception class for reporting all script compile errors in a pipeline.
+ */
+public class PipelineBuildException extends Exception {
```
malaffoon: Please make this extend RuntimeException. I don't really want to get into the checked exception debate but i really don't think any custom exceptions should be checked.

script/src/main/java/org/opentestsystem/rdw/script/ScriptDefinition.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public static class ScriptDefinitionBuilder {
+        private String name;
+        private String version;
+        private ResourceType type;
+        private Class<? extends PipelineScript> compiledScript;
+        private Map<String, Object> configuration;
+        private PropertyResolver propertyResolver;
+
+        ScriptDefinitionBuilder() {
```
malaffoon: shouldn't this be `private`? There's no need for tests to get at it is there?

script/src/main/java/org/opentestsystem/rdw/script/ScriptPipelineConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+import org.opentestsystem.rdw.script.impl.DefaultPropertyResolver;
+import org.opentestsystem.rdw.script.impl.DefaultScriptSource;
+import org.opentestsystem.rdw.script.security.DefaultSandboxPermissions;
+import org.opentestsystem.rdw.script.security.SandboxPermissions;
+import org.opentestsystem.rdw.script.security.SandboxSecurityManager;
+import org.opentestsystem.rdw.script.security.ScriptClassLoader;
+
+/**
+ * Spring configuration for archive service. To use do something like:<pre>
+ * {@literal @}Import(ArchiveConfiguration.class)
```
malaffoon: c&p typo, should be `ScriptPipelineConfiguration`

script/src/main/java/org/opentestsystem/rdw/script/VersionedResource.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Base class for all resources with name and version identifiers.
+ */
+public class VersionedResource {
+    private String name;
+    private String version;
+    private ResourceType type;
+
+    public VersionedResource() {
```
malaffoon: Should this be public?

script/src/main/java/org/opentestsystem/rdw/script/VersionedResource.java
```diff
@@ -0,0 +1,53 @@
+package org.opentestsystem.rdw.script;
+
+import java.util.Objects;
+
+/**
+ * Base class for all resources with name and version identifiers.
+ */
+public class VersionedResource {
+    private String name;
```
malaffoon: Can these be `final`?

script/src/main/java/org/opentestsystem/rdw/script/VersionedResource.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, version, type);
+    }
+
+    @Override
+    public String toString() {
+        return type.name() + ": " + name + ", v. " + version;
```
malaffoon: nit: type isn't important here?

script/src/main/java/org/opentestsystem/rdw/script/impl/DefaultPropertyResolver.java
```diff
@@ -0,0 +1,13 @@
+package org.opentestsystem.rdw.script.impl;
+
+import org.opentestsystem.rdw.script.PropertyResolver;
+
+/**
+ * Default property resolver just returns null (not found) for all properties.
+ */
+public class DefaultPropertyResolver implements PropertyResolver {
```
malaffoon: Maybe a better name would be something like `EmptyPropertyResolver`?

script/src/main/java/org/opentestsystem/rdw/script/security/DefaultSandboxPermissions.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.PropertyPermission;
+import java.util.logging.LoggingPermission;
+
+/**
+ *  Implementation of SandboxPermissions for running user pipeline scripts. The necessary permissions
+ *  are hardcoded in, but this could be changed to load from configuration.
+ */
+public class DefaultSandboxPermissions implements SandboxPermissions {
+    private final @NotNull PermissionCollection whitelist;
+    private final @NotNull PermissionCollection blacklist;
```
malaffoon: nit: i'd either remove the `@NotNull` annotations or explain why they are here?

script/src/main/java/org/opentestsystem/rdw/script/security/SandboxSecurityManager.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Security manager that be enabled or disabled by thread to either check permission against a given set of sandbox
+ * permissions or not check them at all, which would be the default if no security manager is set on the system.
+ * This is more performant than assigning AllPermissions to the application domain and sandbox permissions
+ * to the untrusted domain because there is a big performance penalty for checking permissions at all.
+ */
+public class SandboxSecurityManager extends SecurityManager {
+
+    // Permission needed to disable the security manager.
+    private static final DisableSecurityManagerPermission DISABLE_PERMISSION = new DisableSecurityManagerPermission();
```
malaffoon: Regarding comment from previous review, this is where the DisableSecurityManagerPermission is used (even though you say up above that it is not used). Thus my confusion.

utils/src/test/java/org/opentestsystem/rdw/utils/ParserHelperTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+    @Before
+    public void createParserHelper() {
+        errorCollector = new DataElementErrorCollector();
+        parserHelper = new ParserHelper(errorCollector);
+    }
+
+    @Test
+    public void itShouldValidate() {
+        Assertions.assertThat(parserHelper.validate("field", 23, (Function<Integer, Boolean>) value -> value > 0)).isTrue();
```
malaffoon: Not sure how these got qualified, but can you make them consistent?

script/build.gradle
gregcharles: Hmm, no, it wasn't. I'm sure it's a vestige of something, but I can't remember what.

script/build.gradle
gregcharles: Yes, this one was needed, but not for much. I got rid of it with a couple of test rewrites.

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
gregcharles: There's not a submit call, but I inlined it in the runScript() call and rewrote the recursive call as a loop. Exception reporting is slightly degraded, but it's not an exception I expect to occur often or maybe ever.

script/src/main/java/org/opentestsystem/rdw/script/PipelineBuildException.java
gregcharles: Interesting. I don't think custom exceptions should ever be unchecked. However, since this is probably going away soon anyway, I don't mind making the change. :)

script/src/main/java/org/opentestsystem/rdw/script/security/DefaultSandboxPermissions.java
gregcharles: OK, but it's making me nervous. You had me remove null checks, and now I've removed even the hints that these shouldn't set to null by future developers. They're not null now, but still, it feels a bit sloppy.

script/src/main/java/org/opentestsystem/rdw/script/security/SandboxSecurityManager.java
gregcharles: Ah, I see what you mean. It would be used to disable the SandboxSecurityManager for a thread after it had been enabled. It's that disable functionality that we currently never use. If we removed that, then we'd have no need of the DISABLE_PERMISSION either. 

So, I'm waffling between removing it because we don't need now and leaving it in because we might someday need it and it would be tricky to remember how to  make it work by then. What do you think?

utils/src/test/java/org/opentestsystem/rdw/utils/ParserHelperTest.java
gregcharles: Do you mean the capitalized methods: ItShould...? I also noticed those during the migration. I've changed them now. The method you've commented is already OK by that standard though, so maybe you meant something else?

script/src/main/java/org/opentestsystem/rdw/script/VersionedResource.java
gregcharles: The PipelineDefinition is created from json, and the unmarshaller uses the no args constructor. I could probably work around that with some extra coding. How strongly do you feel about that?

script/src/main/java/org/opentestsystem/rdw/script/VersionedResource.java
gregcharles: Also due to unmarshalling from json. See comment above.

utils/src/test/java/org/opentestsystem/rdw/utils/ParserHelperTest.java
malaffoon: Sorry, wasn't clear ... some of the `assertThat` calls have `Assertions.` and some don't. I'd rather they all don't.

script/src/main/java/org/opentestsystem/rdw/script/VersionedResource.java
malaffoon: In that case, i usually put javadocs explaining why it's there ...
```
/**
 * Default constructor for unmarshaller.
 */
```

PR #103, opened 2019-05-11
Feature/tenant key2

ftdholbrook:
Switch tenant DataSource implementations to use tenant key instead of tenant

- remove dead code
- general clean up and move closer to final production code state
- add description to tenant


multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantProperties.java
```diff
@@ -48,9 +49,11 @@ public void setTenants(List<Tenant> tenants) {
     }
 
     public static class Tenant {
+
         private String id;
         private String key;
         private String displayName;
```
malaffoon: if i asked nicely could we make this just `name`?

multi-tenant/src/test/resources/application-tenant.yml
```diff
@@ -4,6 +4,8 @@ tenant-properties:
     - id: CA
       key: CA
       display-name: California
```
malaffoon: then this could be just `name: California`

multi-tenant/src/test/resources/application-tenant_nv.yml
```diff
@@ -2,7 +2,7 @@
 spring:
   datasource:
     tenants:
-      tenant_nv:
+      NVX:
```
malaffoon: Just reminding myself ... the value here, `NVX`, is the tenant id or key?

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantProperties.java
ftdholbrook: sure

PR #105, opened 2019-05-15
Feature/archive multi tenant

ftdholbrook:
There was an attempt to maintain backwards compatibility, but there is a non-zero chance that I didn't succeed.  I plan to wait until I have reporting converted over before merging this.


archive/src/test/java/org/opentestsystem/rdw/archive/S3ArchiveServiceIT.java
```diff
@@ -24,44 +33,85 @@
...
+ * <li>test-archive.cloud.aws.credentials.secretKey</li>
  * </ul>
  * For example,<pre>
- * java -Dtest-archive.root=s3://rdw-ci-archive \
- *      -Dtest-archive.cloud.aws.credentials.accessKey=MyAccessKey \
- *      -Dtest-archive.cloud.aws.credentials.secretKey=MySecretKey
+ * java -Dtest-archive.uri-root=s3://rdw-ci-archive \
+ *      -Dtest-archive.s3-access-key=MyAccessKey \
+ *      -Dtest-archive.s3-secret-key=MySecretKey
  * </pre>
```
ftdholbrook: If CI is tied into these properties it will need to be modified to match.

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveServiceDecorator.java
```diff
@@ -0,0 +1,72 @@
...
+import org.opentestsystem.rdw.multitenant.archive.ArchiveProperties;
+
+import java.io.InputStream;
+import java.util.Properties;
+
+/**
+ * Decorator to swap between {@link LocalArchiveService} and {@link S3ArchiveService} based
+ * upon the supplied {@link ArchiveProperties}
+ */
+public class ArchiveServiceDecorator implements ArchiveService {
```
malaffoon: Hmm, Decorator doesn't seem right. Facade? Meh, not gonna lose sleep over it.

ftdholbrook: Yeah, I thought facade too, talked with Charlie and looked at the GoF book... Facade warps the underlying implementation, decorator swaps implementation at run time.

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -39,30 +43,52 @@
...
+    public S3ArchiveService(ArchiveProperties archiveProperties) {
+
+        if (!archiveProperties.isConfiguredS3()) {
+            throw new IllegalArgumentException("Invalid S3 root " + archiveProperties.getUriRoot());
         }
+        final AmazonS3ClientBuilder builder = AmazonS3ClientBuilder.standard();
+        builder.setCredentials(new StaticCredentialsProvider(new BasicAWSCredentials(
+                archiveProperties.getS3AccessKey(),
+                archiveProperties.getS3SecretKey())));
+        builder.setRegion(archiveProperties.getS3RegionStatic());
```
malaffoon: I'm a bit concerned that this removes our ability to customize the AmazonS3 client using all the config options available. Can we mimic the Amazon/Spring properties layout and then let Spring autoconfig create the AmazonS3 object?

ftdholbrook: - This is exactly what the code was doing before (except in the factory instead of the constructor)

ftdholbrook:
Just realized I need to add a per-tenant router, leaving this PR open, but do not merge

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
malaffoon: Hmm. Technically that doesn't address my comment, but i guess it is a fair defense  ;)

ftdholbrook: perhaps more salient: the spring auto-configs don't lend themselves to multi tenancy.  this way if we have a future "app global" need we can use the spring configs in parallel.

ftdholbrook:
Updated the code to included a per-tenant router, also reorganized the code into the archive project.  Also, the PR for the reporting side changes are now in place.

PR #106, opened 2019-05-15
Enhance error handling and interfaces to support front-end integration

gregcharles:
Wrap Script.run() in order to wrap script runtime errors in a known type and to capture file and line number information. Added a method to PipelineFactory to ease getting an ad hoc pipeline for testing user scripts in development. Added converters from various exception types to a structure that can be returned to front end.


script/src/main/java/org/opentestsystem/rdw/script/ScriptCompiler.java
```diff
@@ -1,6 +1,7 @@
 package org.opentestsystem.rdw.script;
 
 import groovy.lang.GroovyClassLoader;
+import org.apache.commons.lang3.StringUtils;
```
esotrope: i feel like we usually use spring stringutils instead. @malaffoon ?

script/src/main/java/org/opentestsystem/rdw/script/ScriptRuntimeException.java
```diff
@@ -0,0 +1,65 @@
...
+        Pair<String,Integer> scriptInfo = extractScriptInfo(cause);
+
+        this.scriptFileName = scriptInfo.getLeft();
+        this.lineNumber = scriptInfo.getRight();
+    }
+
+    @Override
+    public String getMessage() {
+        final String scriptName = this.scriptFileName == null ? "UNKNOWN" : this.scriptFileName;
+        final String lineNumber = this.lineNumber == null ? "UNKNOWN" : Integer.toString(this.lineNumber);
```
esotrope: do we need this line if we exit early below when there is no line number?

script/src/main/java/org/opentestsystem/rdw/script/ScriptRuntimeException.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public Integer getLineNumber() {
+        return lineNumber;
+    }
+
+    public String getOriginalMessage() {
+        return super.getMessage();
+    }
+
+    private Pair<String,Integer> extractScriptInfo(final Throwable t) {
```
esotrope: i would prefer something hard-typed

script/src/main/java/org/opentestsystem/rdw/script/util/ExceptionConverter.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Convert a CompilationFailedException, which wraps one or more compilation errors.
+     * @param e the CompilationFailedException
+     * @return List of ScriptErrors, populated with the compile error messages, start lines, and start columns
+     */
+    public static List<ScriptError> convert(final CompilationFailedException e) {
+        if (e instanceof MultipleCompilationErrorsException) {
+            return convert((MultipleCompilationErrorsException)e);
+        } else {
```
esotrope: nit: no need for else block

script/src/main/java/org/opentestsystem/rdw/script/util/ScriptError.java
```diff
@@ -0,0 +1,31 @@
+package org.opentestsystem.rdw.script.util;
+
+import java.io.Serializable;
+
+/**
+ * ScriptError holds basic information about script errors extracted from the
+ * various exceptions thrown from the script compiler and script runtime.
+ */
+public class ScriptError implements Serializable {
```
esotrope: 👍

script/src/main/java/org/opentestsystem/rdw/script/ScriptCompiler.java
```diff
@@ -13,8 +14,38 @@ public ScriptCompiler(GroovyClassLoader classLoader) {
...
+     */
+    public Class<? extends PipelineScript> compile(final String scriptCode, final String name)
+            throws CompilationFailedException {
+        if (StringUtils.isBlank(name)) {
+            return compile(scriptCode);
+        }
+
+        final String fileName = (name.endsWith(".groovy")) ? name : name + ".groovy";
+        Class<?> clazz = classLoader.parseClass(scriptCode, fileName);
+        return clazz.asSubclass(PipelineScript.class);
```
esotrope: clazz. is that when class meets jazz or class meets spaz?

script/src/main/java/org/opentestsystem/rdw/script/ScriptRuntimeException.java
gregcharles: Good point. I meant to take that exit part out.

script/src/main/java/org/opentestsystem/rdw/script/ScriptRuntimeException.java
gregcharles: You'd prefer I define a nonce class to use as the return value for this private method?

script/src/main/java/org/opentestsystem/rdw/script/ScriptCompiler.java
gregcharles: :) Pretty standard naming for an instance of Class though.

gregcharles:
I'm going to merge this now to test out a possible fix to the class loading problem we're seeing in QA. I'll leave the branch open though and add any other corrections to it.

script/src/main/java/org/opentestsystem/rdw/script/ScriptCompiler.java
malaffoon: I like both, pretty sure lang3 has seen some use in the projects.

script/src/main/java/org/opentestsystem/rdw/script/ScriptCompiler.java
malaffoon: Since you're not doing anything with it you could inline it to avoid Landon's taunting  ;)
```
return classLoader.parseClass(scriptCode, fileName).asSubclass(PipelineScript.class);
```

script/src/main/java/org/opentestsystem/rdw/script/ScriptRuntimeException.java
malaffoon: Personally, i'd avoid the whole return thing by inlining it in the constructor. If you really like to break out the helper, let it set the class members directly. If you do leave it like this, this method can be `static`.

script/src/main/java/org/opentestsystem/rdw/script/ScriptRuntimeException.java
```diff
@@ -0,0 +1,61 @@
...
+        if (t.getStackTrace() != null) {
+            for (StackTraceElement element : t.getStackTrace()) {
+                if (StringUtils.endsWith(element.getFileName(), ".groovy")) {
```
malaffoon: Is this doing something you need? That is, why can't you just do
```
    if (element.getFilename().endsWith(".groovy")) {
```
I guess filename might be null? Still think using the utility is confusing
```
    if (element.getFilename() != null && element.getFilename().endsWith(".groovy")) {
```

script/src/main/java/org/opentestsystem/rdw/script/ScriptRuntimeException.java
```diff
@@ -0,0 +1,61 @@
...
+                    final String fileName = element.getFileName();
+                    final Integer lineNumber = (element.getLineNumber() < 0) ? null : element.getLineNumber();
+
+                    return Pair.of(fileName, lineNumber);
+                }
+            }
+        }
+
+        // Not found
+        return Pair.of(null, null);
```
malaffoon: Yeah, this makes me really want to inline this in the constructor.

script/src/main/java/org/opentestsystem/rdw/script/impl/DefaultPipelineFactory.java
```diff
@@ -98,4 +103,43 @@ public Pipeline getPipeline(final String pipelineName, final String pipelineVers
 
         return new Pipeline(pipelineDefinition, scriptDefinitions, sandboxSecurityManager);
     }
+
+    /**
+     * Creates an ad-hoc but runnable pipeline for testing a script during development.
```
malaffoon: I'm not clear on this. Is this a pipeline for unit tests? Or a pipeline that is created until we can load them properly?

script/src/main/java/org/opentestsystem/rdw/script/util/ExceptionConverter.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.opentestsystem.rdw.script.ScriptRuntimeException;
+
+/**
+ * Converters for various script exceptions into ScriptError objects for consumption by front-end services.
+ */
+public class ExceptionConverter {
```
malaffoon: nit: i would've followed the convention of calling this `ScriptErrors`. As it it, the code will look like `ExceptionConverter.convert(e)` which doesn't really tell you anything about what is happening. I think `ScriptErrors.convert(e)` gives a bit of a hint. An alternative is to rename the methods, e.g. `ExceptionConverter.toScriptError(e)`.

script/src/main/java/org/opentestsystem/rdw/script/util/ExceptionConverter.java
```diff
@@ -0,0 +1,85 @@
...
+
+        return Collections.singletonList(convert((GroovyRuntimeException)e));
+    }
+
+    private static List<ScriptError> convert(final MultipleCompilationErrorsException e) {
+        if (e.getErrorCollector() == null
+                || e.getErrorCollector().getErrors() == null
+                || e.getErrorCollector().getErrors().isEmpty())
+        {
+            return Collections.singletonList(convert((GroovyRuntimeException) e));
```
malaffoon: nit: return values with different characteristics can be confusing. If i call this method and get a regular mutable list sometimes and a special immutable list other times makes the calling code potentially complex.

script/src/main/java/org/opentestsystem/rdw/script/ScriptCompiler.java
esotrope: haha. just to be clear i have no bone to pick with the code i just find this naming convention funny. If it were me i would go with `scriptClass` but inlining - that's even better : j

script/src/main/java/org/opentestsystem/rdw/script/util/ExceptionConverter.java
esotrope: `ScriptErrors.toScriptError(e)` would be my vote

PR #107, opened 2019-05-16
emit big scary stack trace just once

malaffoon:
@ftdholbrook, can i do something like this? The stack traces in awsqa make it very difficult to see real problems


ftdholbrook:
well, if it is coming from a different place on the stack trace, it will be hidden.  Maybe a feature toggle instead, so you can turn it off during other testing?

ftdholbrook:
or store the stack trace and hash it and only emit if you see it again... I'm a little worried we have missed some path in the code that is losing the tenant id.

malaffoon:
I figured we'd have to fix the issues. Each time we fix one, the next one would bubble up and give its stack trace.

PR #108, opened 2019-05-23
adds list method to archive service

esotrope:


archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -38,6 +40,7 @@
 public class S3ArchiveService implements ArchiveService {
     private static final Logger logger = LoggerFactory.getLogger(S3ArchiveService.class);
     private static final String S3SchemePrefix = "s3://";
+    private static final String Delimiter = "/";
```
ftdholbrook: Is this a constant anywhere in the S3 lib?  possibly not I guess all the delimiters are just part of the key.

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
```diff
@@ -24,6 +25,16 @@
      */
     String getRawUri(String location);
 
+    /**
+     * Lists all resource paths for or associated with a given location
+     * This can be thought of as listing the contents of a directory.
+     * The returned resource paths are relative - not absolute
+     *
```
malaffoon: Explicit example would be nice. 

Relative to what, the passed in location? That means the caller has to concat the results? Personally, i think it should return a list of locations that could be directly passed into other calls in the service.

archive/src/main/java/org/opentestsystem/rdw/archive/LocalArchiveService.java
```diff
@@ -68,6 +73,17 @@ public String getRawUri(final String location) {
         return FileSchemePrefix + realResourceFile(location).getAbsolutePath();
     }
 
+    @Override
+    public List<String> listResources(final String path) {
+        final String[] files = realResourceFile(path).list();
```
malaffoon: This will return directories as well as files. But directories are not considered resources from the archive service's point of view. I think we should be returning just files. We would have to document the method that that is the behavior: it is just returning real resources with given prefix.

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
malaffoon: That's correct ... technically S3 doesn't care what you put in there, we just use that delimiter by convention.

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -74,20 +77,35 @@ public S3ArchiveService(ArchiveProperties archiveProperties) {
...
     @Override
     public String getRawUri(final String location) {
-        final String prefixedLocation = prefixedLocation(location);
-        return S3SchemePrefix + bucket + "/" + prefixedLocation;
+        return S3SchemePrefix + bucket + Delimiter + prefixedLocation(location);
+    }
+
+    @Override
+    public List<String> listResources(final String locationPrefix) {
+        return amazonS3.listObjectsV2(bucket, prefixedLocation(locationPrefix))
```
malaffoon: Perhaps add a javadoc comment that certain implementations of this method will return only the first N resources (N=1000?)  Or code this to iterate.

archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveService.java
esotrope: i thought when doing lookups we prepend the bucket & some prefix for them so relative paths to that is what we want to return. but i'll check

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
esotrope: ah good point. the s3 method pages huh. hm...

archive/src/main/java/org/opentestsystem/rdw/archive/S3ArchiveService.java
```diff
@@ -38,56 +44,104 @@
...
     public String getRawUri(final String location) {
-        final String prefixedLocation = prefixedLocation(location);
-        return S3SchemePrefix + bucket + "/" + prefixedLocation;
+        return S3SchemePrefix + bucket + Delimiter + prefixedLocation(location);
+    }
+
+    @Override
+    public List<String> listResources(final String locationPrefix) {
+        final List<String> resourcePaths = new ArrayList<>();
+        try {
```
esotrope: now makes continual calls until all keys are pulled

PR #109, opened 2019-05-23
RP-387: fix security hole with scripts starting new threads

gregcharles:
I tried a few different fixes for this. One was to default the security manager to be enabled for new threads and then just disable it for the main thread. That worked to an extent, but some non-script threads still ended up with an enabled security manager, and in some of the JUnits, this caused a deadlock in a finalizer. I couldn't figure out exactly why. 

In any case, this solution, when it finally occurred to me is laughably simple, and yet it's effective. It might be hackable, but I couldn't figure out a way.

### DIFF
--- a/script/src/main/resources/scripts/DSLScriptBase.groovy
+++ b/script/src/main/resources/scripts/DSLScriptBase.groovy
@@ -26,6 +26,10 @@ import java.nio.charset.StandardCharsets
  * Base class for Groovy pipeline scripts. Provides a DSL that will be available to the scripts.
  */
 abstract class DSLScriptBase extends PipelineScript {
+    static {
+        Thread.metaClass.start {throw new SecurityException("Illegal to start new thread from script")}
+    }
+
     // Collection of errors found by validating scripts.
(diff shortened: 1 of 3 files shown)


script/src/main/resources/scripts/DSLScriptBase.groovy
```diff
@@ -26,6 +26,10 @@ import java.nio.charset.StandardCharsets
  * Base class for Groovy pipeline scripts. Provides a DSL that will be available to the scripts.
  */
 abstract class DSLScriptBase extends PipelineScript {
+    static {
+        Thread.metaClass.start {throw new SecurityException("Illegal to start new thread from script")}
```
esotrope: this is like the Thread class prototype? that is simple

esotrope: can someone in the script write to `metaClasses`?

malaffoon: A code comment would be nice, something like
```
    // Threads have separate security so we don't want to let a script create one.
    // This seems to be the simplest way to stop that from happening.
```

PR #112, opened 2019-05-24
add message header for tenant id

malaffoon:
On the RDW_Ingest side most services don't know anything about users so we just need to get the tenant id in the message.

This is a pretty trivial change so i'll push it and address any comments later.


multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantProperties.java
```diff
@@ -5,8 +5,13 @@
 import java.util.Optional;
```
malaffoon: unrelated, just adding javadocs cause i spend a few minutes discovering the info

PR #113, opened 2019-05-24
RP-395 Provide some level of visibility around what tenant configurat…

ftdholbrook:
…ions are active, and for how long they have been active.

### DIFF
--- a/multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantConfigurationStatus.java
+++ b/multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantConfigurationStatus.java
@@ -1,0 +1,71 @@
+package org.opentestsystem.rdw.multitenant;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import java.util.List;
+
+import java.time.LocalDateTime;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public class TenantConfigurationStatus {
+    LocalDateTime created;
+    List<String> tenantKeys;
+    String additionalContext;
+
+    public LocalDateTime getCreated() {
+        return created;
+    }
+
+    public void setCreated(LocalDateTime created) {
+        this.created = created;
+    }
+
+    public List<String> getTenantKeys() {
+        return tenantKeys;
+    }
+
+    public void setTenantKeys(List<String> tenantKeys) {
+        this.tenantKeys = tenantKeys;
+    }
+
+    public String getAdditionalContext() {
+        return additionalContext;
+    }
+
+    public void setAdditionalContext(String additionalContext) {
+        this.additionalContext = additionalContext;
+    }
+
+
+    public static final class TenantConfigurationStatusBuilder {
+        private TenantConfigurationStatus tenantConfigurationStatus;
+
+        private TenantConfigurationStatusBuilder() {
+            tenantConfigurationStatus = new TenantConfigurationStatus();
+        }
+
+        public static TenantConfigurationStatusBuilder aTenantConfigurationStatus() {
+            return new TenantConfigurationStatusBuilder();
+        }
+
+        public TenantConfigurationStatusBuilder withCreated(LocalDateTime created) {
+            tenantConfigurationStatus.setCreated(created);
+            return this;
+        }
+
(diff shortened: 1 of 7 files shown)


multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantConfigurationStatus.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.util.List;
+
+import java.time.LocalDateTime;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public class TenantConfigurationStatus {
+    LocalDateTime created;
+    List<String> tenantKeys;
+    String additionalContext;
```
malaffoon: `private`

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantConfigurationStatus.java
```diff
@@ -0,0 +1,71 @@
...
+    public String getAdditionalContext() {
+        return additionalContext;
+    }
+
+    public void setAdditionalContext(String additionalContext) {
+        this.additionalContext = additionalContext;
+    }
+
+
+    public static final class TenantConfigurationStatusBuilder {
```
malaffoon: Some conventions for RDW builder pattern:
* inner builder class is called `Builder`
* setters do not use `with`, `a`, etc.; example for this one, setter would be `created` not `withCreated`
* builder instance is retrieved using a static `builder()` method on the built class
* setters on built class are either not there (and builder just directly sets members) or private

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantConfigurationStatus.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.time.LocalDateTime;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public class TenantConfigurationStatus {
+    LocalDateTime created;
+    List<String> tenantKeys;
+    String additionalContext;
+
+    public LocalDateTime getCreated() {
```
malaffoon: Not a hill i'll die on but just wanted to ask if you thought about LocalDateTime vs. Instant? Obviously, LocalDateTime is a bit easier to grok but it will be harder to correlate to database timestamps which are all GMT instants.

ftdholbrook: Not doing any fine grained timing with this, and it auto-serialize to a readable format.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantConfigurationStatus.java
ftdholbrook: I just used the intelij plugin.  I need public setters or whatever serializer used by the info plugin barfs (tried it).  I can just remove this and go back to a plain bean if it is confusing.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantConfigurationStatus.java
ftdholbrook: default privates in scala spoiled me.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantConfigurationStatus.java
ftdholbrook: ... and I don't expect this ever to get stored in a db

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantConfigurationStatus.java
malaffoon: We are deserializing this? I was assuming it was just being serialized, not deserialized.

I like builders though.

ftdholbrook: tried removing them and it worked this time.. was having earlier problems with actuator, so must have been unrelated

PR #114, opened 2019-05-24
RD-386: infrastructure changes to support saving transformed docs.

gregcharles:
These library changes support the suggestions made for the rp-386 pr in RDW_Ingest. Regarding the updates to LocationStrategy, I'm assuming we _will_ recompute location from the digest.

### DIFF
--- a/script/src/test/java/org/opentestsystem/rdw/script/XmlPipelineTest.java
+++ b/script/src/test/java/org/opentestsystem/rdw/script/XmlPipelineTest.java
@@ -1,5 +1,6 @@
 package org.opentestsystem.rdw.script;
 
+import org.apache.commons.lang3.RandomUtils;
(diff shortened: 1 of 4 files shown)


script/src/test/java/org/opentestsystem/rdw/script/XmlPipelineTest.java
```diff
@@ -201,6 +205,20 @@ public void itShouldFilterInElements() throws Exception {
         assertThat(newXml, hasXPath("//Item[not(@key='37115')]", isBlankString()));
     }
 
+    @Test
+    public void isShouldReportNoModifications() {
+        final byte [] input = RandomUtils.nextBytes(1000);
```
malaffoon: interesting, didn't want to just do `final byte[] input = "Some input".getBytes();`?

PR #115, opened 2019-05-24
Feature/common script publishing

esotrope:


script/src/main/java/org/opentestsystem/rdw/script/PipelineDefinition.java
```diff
@@ -10,18 +10,18 @@
 /**
  * Definition of a Pipeline including its name, version, node definitions, and roll back information.
  */
-public class PipelineDefinition extends VersionedResource {
+public class PipelineDefinition extends AbstractPipelineResource {
     private String rollbackVersion;
```
gregcharles: Weren't you going to remove rollbackVersion?

esotrope: not quite there yet. - had to take a step back remember

PR #120, opened 2019-05-28
adds guard for reading pipeline resource

esotrope:


script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -98,10 +98,14 @@ public ArchiveBackedPublishedPipelineRepository(
...
+
+        if (!archiveService.exists(path)) {
+            return null;
+        }
+
         final PipelineResource resource = deserialize(
-                archiveService.readResource(
-                        toPipelinePath(code, version)
-                ),
+                archiveService.readResource(path),
```
esotrope: oops. readResource throws if the resource is absent

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -98,10 +98,14 @@ public ArchiveBackedPublishedPipelineRepository(
...
                 PipelineResource.class
         );
```
gregcharles: Would this take two calls to the backing archive? Could you just handle the exception and save the extra call?

esotrope: i could. fishing for permission to update the archive service api to return an optional first though ; J

malaffoon: I'm fine with that. You'll need to update all usage of course.

esotrope: 👍 i'll make a jira

PR #122, opened 2019-05-28
adds better configuration validation to published pipelines

esotrope:


script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -109,11 +108,31 @@ public PublishedPipeline findByCodeAndVersion(final String code, final String ve
...
+                PublishedScript.builder()
+                        .type(Pre)
+                        .version(resource.getPreProcessScriptVersion())
+                        .build(),
+                PublishedScript.builder()
+                        .type(Post)
+                        .version(resource.getPostProcessScriptVersion())
+                        .build()
+        ).stream()
+                .filter(script -> script.getVersion() != null)
```
esotrope: so if the version isn't set in the def don't even look for the script.

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -131,27 +150,41 @@ public PublishedPipeline create(final PublishedPipeline pipeline) {
...
         );
 
         final PublishedPipeline published = pipeline.copy()
                 .version(nextVersion)
                 .published(Instant.now())
+                .scripts(
+                        pipeline.getScripts().stream()
+                                .map(script ->
+                                        script.getType() == User
+                                                ? script.copy().version(nextVersion).build()
```
esotrope: give the user script the same version as the pipeline

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -176,30 +209,35 @@ public PublishedPipeline create(final PublishedPipeline pipeline) {
...
+            final ScriptType type,
+            final String version) {
+
+        final String scriptPath = toScriptPath(type, pipelineCode, version);
+        if (archiveService.exists(scriptPath)) {
+            return readResourceAsString(scriptPath);
         }
-        return Optional.empty();
+        final String pipelinePath = toPipelinePath(pipelineCode, pipelineVersion);
+        throw new IllegalStateException(String.format(
```
esotrope: catches the situation where we have a `xyzScriptVersion: N` entry in the `pipeline.json` but the file isn't actually there.

script/src/main/java/org/opentestsystem/rdw/script/publishing/PipelineResource.java
```diff
@@ -48,6 +53,7 @@ public static Builder builder() {
         private String version;
         private Instant published;
         private String publishedBy;
+        private String userScriptVersion;
```
esotrope: the value for this will always be the same as version but i figured this was more straight forward and explicit

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -131,27 +150,41 @@ public PublishedPipeline create(final PublishedPipeline pipeline) {
                 findAllByCode(pipeline.getPipelineCode())
                         .stream()
                         .mapToInt(published -> Integer.parseInt(published.getVersion()))
-                        .max().orElse(0) + 1
+                        .max()
+                        .orElse(0)
+                        + 1
```
malaffoon: personally i think one liner is easier to read

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
malaffoon: Okay. Why?

script/src/main/java/org/opentestsystem/rdw/script/publishing/PipelineResource.java
malaffoon: clearly a conversation was had ...

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
esotrope: i try to keep every `.xyz()` of a chain on its own line for readability

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
esotrope: otherwise the consumer of this interface would need to compute the next ID for the script. have a different solution in mind?

malaffoon: Just curious if it is valid and correct to tie the pipeline version to the user script version.

esotrope: ah. yeah. they both increment in lock step on publish. because you are only able to publish when there is a new script. maybe it would be prudent for us to find the "next version" of the script too for times when there is manual provisioning going on i guess though huh..

malaffoon:
@esotrope can we close this PR?

esotrope:
@malaffoon closing. I have totally forgotten the context for this so it is probably not relevant anymore

PR #125, opened 2019-05-30
removes archive service declaration to avoid duplicate beans

esotrope:


script/src/main/java/org/opentestsystem/rdw/script/ScriptPipelineConfiguration.java
```diff
@@ -15,79 +11,74 @@
...
+ *     this configuration as it depends on an archive service to operate
  *
  * @see PipelineFactory
  * @see ScriptCompiler
  * @see GroovyClassLoader
  */
 @Configuration
-@Import({
-        ArchiveConfiguration.class
-})
```
esotrope: the only real change

PR #126, opened 2019-05-30
adds exception safety back

esotrope:


script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -57,13 +57,14 @@ public ArchiveBackedPublishedPipelineRepository(
 
     @Override
     public PublishedPipeline findByCodeAndVersion(final String code, final String version) {
-        final Properties properties = findUserScriptProperties(code, version);
-        if (properties == null) {
+        try {
```
esotrope: got confused and thought i could remove this guard but it is only because i forgot to mock the archive service to throw which is what it does irl

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -57,13 +57,14 @@ public ArchiveBackedPublishedPipelineRepository(
...
+            return toPublishedPipeline(findUserScriptProperties(code, version)).copy()
+                    .scripts(findScripts(code, version))
+                    .build();
+        } catch (final Exception exception) {
```
malaffoon: This should be `IllegalArgumentException`

PR #128, opened 2019-05-30
updates common-script to support reporting admin needs

esotrope:


script/src/main/java/org/opentestsystem/rdw/script/PipelineFactory.java
```diff
@@ -6,6 +6,11 @@
  */
 public interface PipelineFactory {
 
+    /**
+     * TODO
+     */
+    Pipeline getEffectivePipeline(String pipelineCode, String pipelineVersion) throws PipelineBuildException;
```
esotrope: creates a pipeline regardless of whether there is one published or not - does so by getting the base and system scripts using the provided pipeline code

script/src/main/java/org/opentestsystem/rdw/script/impl/DefaultPipelineFactory.java
```diff
@@ -54,37 +59,116 @@ private static PipelineDefinition toPipelineDefinition(final PublishedPipeline p
...
+                                .map(PublishedScript::getBody)
+                                .orElse(null)
+                        )
+                        .build()
+                )
+                .filter(script -> script.getBody() != null)
+                .collect(toList());
+    }
+
+    private Pipeline createPipeline(final PublishedPipeline publishedPipeline) throws PipelineBuildException {
```
esotrope: found that the getPipeline and getDeveloperPipeline could both use this base code 👍

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -48,7 +46,7 @@ public ArchiveBackedPublishedPipelineRepository(
     @Override
     public List<PublishedPipeline> findAllByCode(final String code) {
         return archiveService.listResources(toPipelinesPathPrefix(code)).stream()
-                .filter(path -> UserScriptPattern.matcher(path).find())
+                .filter(path -> path.endsWith(".groovy"))
```
esotrope: didn't realize the properties were stored as files with a pattern that matched the user script pattern.

gregcharles: Yes, that's true for the local archive. In s3, the properties are stored as metadata on the file.

script/src/main/java/org/opentestsystem/rdw/script/PipelineFactory.java
gregcharles: Do we need another interface? How is this different from the regular getPipeline()? This change will break ingest until it's reintegrated.

esotrope: will only break if we bump the version of script-common on ingest.
what this does is allow you to get a pipeline regardless of whether a userscript is published or not. the current implementation throws an error if there is no user script for the requested pipeline. an NPE i think. the method impl of this creates a pipeline with just the base and system scripts if the user script is absent

malaffoon: It feels like we should be changing getPipeline to have that logic, somehow. It will be confusing to have multiple factory methods like this.

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
malaffoon: It feels like we should be matching the filenames even more precisely. Kind of like you had it before?

I also think we need more javadocs or code comments about the expected behavior.

PR #129, opened 2019-05-30
Task/effective pipeline creation

esotrope:


script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
```diff
@@ -47,6 +50,10 @@ public PipelineDefinition getPipelineDefinition() {
         return pipelineDefinition;
     }
 
+    public List<ScriptDefinition> getScripts() {
+        return scriptDefinitions;
```
esotrope: exposed this for testability.

@gregcharles one thing i noticed is we can simplify things to not use the "PipelineDefinition" at all i think. right now we map published pipeline to pipeline def and then never use what is on the pipeline def - only what is passed here as script defs. is that a correct observation?

script/src/main/java/org/opentestsystem/rdw/script/impl/DefaultPipelineFactory.java
```diff
@@ -75,67 +75,56 @@ public DefaultPipelineFactory(final PublishedPipelineRepository repository,
...
-                                                "No base script found for the \"%s\" pipeline",
-                                                pipelineCode
-                                        )))
-                        ))
-                        .build()
+                createDevelopmentPipeline(pipelineCode, userScript)
         );
     }
 
+    private PublishedPipeline getEffectivePipeline(
```
esotrope: new logic block for getting the effective pipeline "def"

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -33,20 +30,18 @@
 @Repository
 public class ArchiveBackedPublishedPipelineRepository implements PublishedPipelineRepository {
 
+    private static final Pattern UserScriptPattern = Pattern.compile("^(.*?)user\\.(\\d+)\\.groovy$");
```
esotrope: _more_ exact than `.groovy$` but is it exact enough @malaffoon ?

script/src/main/java/org/opentestsystem/rdw/script/publishing/PublishedPipelines.java
```diff
@@ -8,6 +8,12 @@
  */
 public final class PublishedPipelines {
 
+    // Note the property keys had to be all lower case due to the way Amazon S3 coerces all property names to lower case
+    private static final String PipelineCodeProperty = "pipeline_code";
+    private static final String VersionProperty = "version";
+    private static final String PublishedOnProperty = "published_on";
+    private static final String PublishedByProperty = "published_by";
```
esotrope: lower case props

script/src/test/java/org/opentestsystem/rdw/script/PipelineFactoryTest.java
```diff
@@ -71,54 +93,148 @@ public void before() throws Exception {
...
+        factory.getPipeline(publishedPipeline.getPipelineCode(), null);
+    }
+
+    @Test(expected = PipelineBuildException.class)
+    public void getPipelineShouldThrowExceptionWhenPostScriptDoesntCompile() {
+        when(scriptSource.loadScript(Post, publishedPipeline.getPipelineCode()))
+                .thenReturn(Optional.of(NonCompilingScript));
+
+        factory.getPipeline(publishedPipeline.getPipelineCode(), null);
+    }
```
esotrope: built out a bunch more tests for the pipeline factory - still need more

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
malaffoon: Looks fine. Why the groupings, do we need them?

esotrope: nope - will remove

PR #130, opened 2019-05-31
Task/simplify common script and add coverage

esotrope:
next pass i'll be adding a bunch of coverage


script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
```diff
@@ -28,15 +27,15 @@
     private static final int SANDBOX_THREADS = 2;
     private static final ExecutorService THREAD_POOL = Executors.newFixedThreadPool(SANDBOX_THREADS);
 
-    private final PipelineDefinition pipelineDefinition;
-    private final List<ScriptDefinition> scriptDefinitions;
+    private final PipelineDefinition source;
+    private final List<PipelineScript> scripts;
```
esotrope: really the pipeline doesn't even need it's source but i figured i would keep the pattern going.
also i thought it made more sense fore the pipeline scripts to be instantiated in the factory - or do we consider instantiation issues runtime errors? i could go either way on it so i can revert if there is a strong opinion.

script/src/main/java/org/opentestsystem/rdw/script/PipelineDefinition.java
```diff
@@ -1,62 +1,117 @@
...
 import java.util.List;
 
 import static com.google.common.collect.ImmutableList.copyOf;
 import static com.google.common.collect.ImmutableList.of;
 
 /**
- * Definition of a Pipeline including its name, version, node definitions, and roll back information.
+ * Represents a published pipeline artifact
  */
 public class PipelineDefinition {
```
esotrope: what happened here is i deleted `PipelineDefinition` and renamed `PublishedPipeline` to `PipelineDefinition`
Also, I renamed `PublishedScript` to `PipelineScriptDefinition`

script/src/main/java/org/opentestsystem/rdw/script/ScriptRuntimeException.java
```diff
@@ -5,22 +5,30 @@
...
-        final String scriptName = this.scriptFileName == null ? "UNKNOWN" : this.scriptFileName;
-        final String lineNumber = this.lineNumber == null ? "UNKNOWN" : Integer.toString(this.lineNumber);
-
-        return "Error in " + scriptName + " at line " + lineNumber + ": " + getOriginalMessage();
+        return String.format(
+                "Error in %s at line %s: %s",
+                this.scriptFileName == null ? "UNKNOWN" : this.scriptFileName,
+                this.lineNumber == null ? "UNKNOWN" : Integer.toString(this.lineNumber),
+                getOriginalMessage()
+        );
```
esotrope: applied code styles in some files

script/src/main/java/org/opentestsystem/rdw/script/impl/DefaultPipelineFactory.java
```diff
@@ -1,61 +1,40 @@
...
+import static org.opentestsystem.rdw.script.PipelineScriptType.Pre;
+import static org.opentestsystem.rdw.script.PipelineScriptType.User;
 
 /**
  * Factory class for creating a Pipeline by loading the pipeline definition and then using it to
  * load the script definitions for each node.
  */
 public class DefaultPipelineFactory implements PipelineFactory {
 
-    private static PipelineDefinition toPipelineDefinition(final PublishedPipeline publishedPipeline) {
```
esotrope: this transform is no longer needed yay

script/src/test/java/org/opentestsystem/rdw/script/PipelineDefinitionJsonTest.java
```diff
@@ -1,51 +1,14 @@
...
-import org.opentestsystem.rdw.script.publishing.PublishedPipeline;
-import org.opentestsystem.rdw.script.publishing.PublishedScript;
-import org.opentestsystem.rdw.script.publishing.ScriptType;
-
-import static com.google.common.collect.Lists.newArrayList;
-import static org.assertj.core.api.Assertions.assertThat;
 
+// DELETE AT YOUR OWN PERIL
+// Ask Greg Charles
 public class PipelineDefinitionJsonTest {
```
esotrope: deleting this causes other tests to fail : J

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
```diff
@@ -28,15 +27,15 @@
...
 
-    public Pipeline(final PipelineDefinition pipelineDefinition,
-                    final List<ScriptDefinition> scriptDefinitions,
+    public Pipeline(final PipelineDefinition source,
+                    final List<PipelineScript> scripts,
                     final SandboxSecurityManager sandboxSecurityManager) {
-        this.pipelineDefinition = pipelineDefinition;
-        this.scriptDefinitions = scriptDefinitions != null ? copyOf(scriptDefinitions) : of();
+        this.source = source;
+        this.scripts = scripts != null ? copyOf(scripts) : of();
```
malaffoon: i would suggest that a collection parameter never be null; so this can always just copy it?

script/src/main/java/org/opentestsystem/rdw/script/PipelineDefinition.java
malaffoon: Brain asplodes

script/src/test/java/org/opentestsystem/rdw/script/PipelineDefinitionJsonTest.java
esotrope: @gregcharles this will be a fun one to troubleshoot - let me know when you have some time to look at this with me

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
esotrope: yeah we could enforce that I suppose

PR #131, opened 2019-05-31
Task/simplify common script and add coverage

esotrope:
added more coverage

### DIFF
--- a/script/src/main/java/org/opentestsystem/rdw/script/security/AbstractToggleableSecurityManager.java
+++ b/script/src/main/java/org/opentestsystem/rdw/script/security/AbstractToggleableSecurityManager.java
@@ -3,7 +3,7 @@ package org.opentestsystem.rdw.script.security;
 /**
  * An abstract security toggleable security manager
  */
-public abstract class SandboxSecurityManager extends SecurityManager {
+public abstract class AbstractToggleableSecurityManager extends SecurityManager {
 
     /**
      * @return True if the security manager is enabled
--- a/script/src/test/java/org/opentestsystem/rdw/script/PipelineTest.java
+++ b/script/src/test/java/org/opentestsystem/rdw/script/PipelineTest.java
@@ -1,33 +1,180 @@
 package org.opentestsystem.rdw.script;
 
 import org.junit.Test;
+import org.opentestsystem.rdw.script.security.AbstractToggleableSecurityManager;
+import org.opentestsystem.rdw.script.security.DefaultSandboxPermissions;
+import org.opentestsystem.rdw.script.security.DefaultSandboxSecurityManager;
+
+import java.util.Collections;
+import java.util.function.Supplier;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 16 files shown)


script/src/main/java/org/opentestsystem/rdw/script/security/AbstractToggleableSecurityManager.java
```diff
@@ -0,0 +1,22 @@
...
+
+    /**
+     * Enables the security manager's security restrictions
+     */
+    public abstract void enable();
+
+    /**
+     * Disables the security manager's security restrictions
+     */
+    public abstract void disable();
```
esotrope: made this an abstract in-between to so it could be mocked better later. unfortunately SecurityManager is not an interface

script/src/test/java/org/opentestsystem/rdw/script/PipelineTest.java
```diff
@@ -1,33 +1,180 @@
...
+    public void constructorShouldThrowExceptionWhenScriptsAreNull() {
+        new Pipeline(
+                pipelineWithoutUserScript,
+                null,
+                securityManager
+        );
+    }
+
+    @Test
+    public synchronized void constructorShouldSetSecurityManagerWhenUserScriptsArePresent() {
```
esotrope: is this sufficient to make sure junit doesn't run this and the next test in parallel?

script/src/test/java/org/opentestsystem/rdw/script/util/Tests.java
```diff
@@ -16,17 +20,40 @@ public static InputStream resourceAsStream(final String path) {
...
-                resourceAsStream(path)
-        );
+    public static byte[] resourceAsBytes(final String path) {
+        try {
+            return IOUtils.toByteArray(
+                    resourceAsStream(path)
+            );
+        } catch (final IOException exception) {
+            throw new RuntimeException(exception);
+        }
```
esotrope: is there a functional way to decorate this method with this kind of exception handling?

script/src/test/java/org/opentestsystem/rdw/script/PipelineTest.java
malaffoon: junit will only run tests in parallel if we configure it to ... which we don't; were you seeing something weird?

malaffoon: And this are independent synchronized blocks. Nothing stops two separate threads from running each test at the same time. Guess i'm not sure what you ran into and what you are trying to avoid?

script/src/test/java/org/opentestsystem/rdw/script/util/Tests.java
malaffoon: nothing comes to mind

script/src/test/java/org/opentestsystem/rdw/script/PipelineTest.java
esotrope: yeah that's what i figured. this didn't look right to me either. i think all i ran into was the tests being run in an order that failed them. but that was fixed by resetting the security manager at the start of the method

PR #132, opened 2019-06-01
RP-431: implement script helpers for CSV input 

gregcharles:
Updates to script base class to support CSV processing


script/src/main/resources/scripts/DSLScriptBase.groovy
```diff
@@ -181,19 +187,135 @@ abstract class DSLScriptBase extends PipelineScript {
...
+
+    //  Adds CSV parsing and transformation helper methods to the script class.
+    //  See readme for specific examples of using these helper methods.
+    private void enableCsvExtensions() {
+        checkExtension('csv')
+
+        def input = getProperty("input")
+
+        if (input == null) {
+            throw new RuntimeException(" CSV Input cannot be null")
```
malaffoon: nit: leading blank in message

script/src/main/resources/scripts/DSLScriptBase.groovy
```diff
@@ -181,19 +187,135 @@ abstract class DSLScriptBase extends PipelineScript {
...
+
+        if (input instanceof String) {
+            reader = new StringReader(input)
+        } else if (input instanceof byte[]) {
+            reader = new InputStreamReader(new ByteArrayInputStream(input), UTF_8)
+        } else if (input instanceof InputStream) {
+            reader = new InputStreamReader(input, UTF_8)
+        } else if (input instanceof Reader) {
+            reader = input
+        } else {
```
malaffoon: Why not put the `input == null` check in with these? Imagine a helper method `final Reader reader = checkInput(input);`

script/src/main/resources/scripts/DSLScriptBase.groovy
```diff
@@ -181,19 +187,135 @@ abstract class DSLScriptBase extends PipelineScript {
...
+            throw new RuntimeException("Unsupported input type: " + input.getClass())
+        }
+
+        CSVFormat csvFormat = CSVFormat.RFC4180.withFirstRecordAsHeader().withCommentMarker('#' as char)
```
malaffoon: might want to add `.withIgnoreHeaderCase()`

gregcharles: I'll keep this in mind, although I duplicated this format from the actual assessment processor. If we did allow ignoring header case, I'd need to experiment with how well it would play with addressing cells by row.header_name.

PR #133, opened 2019-06-01
start locking down no-tenant handling

malaffoon:
Goal is to remove any special consideration of no-tenant, and clean up any non-tenant-aware stuff. I also moved a couple things into common from elsewhere since they are used by both RDW_Ingest and RDW_Reporting.

There are changes in both RDW_Ingest and RDW_Reporting (https://github.com/SmarterApp/RDW_Reporting/pull/1959) that have to be coordinated with this.


archive/src/main/java/org/opentestsystem/rdw/archive/ArchiveServiceTenantDynamicRouter.java
```diff
@@ -35,8 +36,11 @@ protected ArchiveService archiveServiceFactory(String key) {
     }
 
     private ArchiveService getArchiveService() {
-        final String tenantKey = this.tenantKeyResolver.getTenantKey().orElse(TenantKeyResolver.NO_TENANT_KEY);
-        return tenantArchiveServices.computeIfAbsent(tenantKey, this::archiveServiceFactory);
+        final Optional<String> tenantKey = this.tenantKeyResolver.getTenantKey();
+        if (!tenantKey.isPresent()) {
+            throw new IllegalStateException("No tenant key set");
+        }
```
malaffoon: The general philosophy i've adopted is: Resolvers return Optionals, and the thing that uses the Resolver (router, interceptor, etc.) is what throws the exception.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/messaging/TenantMessageRequestResolver.java
```diff
@@ -2,17 +2,16 @@
...
     /**
-     * Resolves the tenant id
-     *
-     * @param message Message
-     * @return tenantId
+     * @param message message from which to extract the tenant id
+     * @return the tenant id
      */
-    String resolveTenantId(Message<?> message);
+    Optional<String> getTenantId(Message<?> message);
```
malaffoon: changed to match overall philosophy

script/src/main/java/org/opentestsystem/rdw/script/publishing/ArchiveBackedPublishedPipelineRepository.java
```diff
@@ -63,7 +63,7 @@ public PipelineDefinition findByCodeAndVersion(final String code, final String v
                                     .build()
                     ))
                     .build();
-        } catch (final Exception exception) {
+        } catch (final IllegalArgumentException ignored) {
```
malaffoon: @esotrope, i mentioned this in one of your reviews; i'm going on the assumption that it is a good change.

esotrope: i changed something to illegal argument exception - cant remember what at this point though x l

PR #134, opened 2019-06-03
Move base script out of package to fix (?) class loading

gregcharles:
This is just for you to try out to see if makes that test failure go away.


script/src/test/java/org/opentestsystem/rdw/script/CsvScriptTest.java
```diff
@@ -14,7 +14,7 @@
 
     @BeforeClass
     public static void setUp() throws Exception {
-        compiler = ScriptCompilers.toScriptCompiler(resourceAsString("/scripts/DSLScriptBase.groovy"));
+        compiler = ScriptCompilers.toScriptCompiler(resourceAsString("/DSLScriptBase.groovy"));
```
esotrope: i'd be tempted to rename this `UserScriptBase.groovy` it at least would help me know that it will become the base class to the user script

script/src/test/java/org/opentestsystem/rdw/script/ScriptCompilerTest.java
```diff
@@ -28,7 +28,7 @@
     @Before
     public void setup() throws Exception {
         compiler = ScriptCompilers.toScriptCompiler(
-                IOUtils.toString(this.getClass().getResourceAsStream("/scripts/DSLScriptBase.groovy"), UTF_8)
+                IOUtils.toString(this.getClass().getResourceAsStream("/DSLScriptBase.groovy"), UTF_8)
```
esotrope: these could use the `resourceAsString` util method

esotrope: actually i think i made a `baseScriptBody()` method to avoid writing the string path constant out everywhere

PR #135, opened 2019-06-07
Adding tenant isSandbox flag and supporting methods

burzum619:


multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/Tenant.java
```diff
@@ -10,11 +12,14 @@
     private String key;
     private String name;
     private String description;
+    private boolean sandbox;
+    private Instant createdAt;
```
esotrope: in rdw the convention is to just name this field `created`

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantProperties.java
```diff
@@ -35,12 +33,34 @@ public TenantProperties() {
...
+                .filter(tenant -> !tenant.isSandbox())
+                .collect(Collectors.toSet());
+    }
+
+    /**
+     * Fetches a collection of all sandboxes
+     *
+     * @return a {@link Collection} of all tenants that are sandboxes
+     */
+    public Set<Tenant> findSandboxes() {
```
esotrope: hm. maybe we have `getTenants()` `getSandboxes()` and `getTenantsAndSandboxes()`? `getTenants()` and `findTenants()` behaving the way they do is a bit tricky

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/Tenant.java
```diff
@@ -28,6 +33,7 @@ public Tenant(final String id, final String key, final String name) {
         this.id = id;
         this.key = key;
         this.name = name;
+        this.createdAt = Instant.now();
```
esotrope: will this eventually be read from yaml?

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantProperties.java
```diff
@@ -4,9 +4,7 @@
 import org.slf4j.LoggerFactory;
 
 import java.time.LocalDateTime;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
malaffoon: nit: splats bad

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantProperties.java
malaffoon: We could just let the caller of `getTenants()` do filtering.

ftdholbrook: for reflection reasons can't use get (this class is populated by @ConfigProperties) there are already some helper finders on here that do similar thing.  It will be used in enough places I think it makes sense to leave it here.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/Tenant.java
ftdholbrook: yes this is what is in the yml file*-

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/TenantProperties.java
esotrope: ah. yeah i'm leaning towards marks inclination. since this is just the pojo for the config file it makes sense for filtering to be in a service or repo

PR #137, opened 2019-06-13
Add schema search path to data source elements

ftdholbrook:
Needed for Redshift tenant creation

### DIFF
--- a/multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/datasource/DataSourceElementsProperties.java
+++ b/multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/datasource/DataSourceElementsProperties.java
@@ -24,6 +24,7 @@ public class DataSourceElementsProperties implements DataSourceElements, TenantC
     private String jdbcInterceptors;
     private String username;
     private String password;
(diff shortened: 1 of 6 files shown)


multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/datasource/DataSourceElementsProperties.java
```diff
@@ -24,6 +24,7 @@
     private String jdbcInterceptors;
     private String username;
     private String password;
+    private String schemaSearchPath;
```
vgonzalez9: So is the what needs to be validated as existing? And being a path do we need to link the path with the db for a validation?

multi-tenant/src/test/resources/application-tenant_ca.yml
```diff
@@ -15,6 +15,7 @@ spring:
           database: "tenant_ca"
         username: "tenant_ca_user"
         password: "tenant_ca_password"
+        schema-search-path: "ca_schema"
```
esotrope: so this is just an extra redshift specific db param i take it?

PR #142, opened 2019-06-19
RP-494_Dataset_Configuration_common

vgonzalez9:
Added DataSets properties of the preconfigured Datasets that are used in creating Sandboxes.

Once built, will update the RDW_Reporting's multitenant lib version.


multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/SandboxDataSets.java
```diff
@@ -0,0 +1,21 @@
+package org.opentestsystem.rdw.multitenant;
+
+import java.util.List;
+
+/**
+ * Container for the collection of Sandbox DataSets
+ */
+public class SandboxDataSets {
```
malaffoon: Are we really going to need a container for the collection?

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/SandboxDataSets.java
```diff
@@ -0,0 +1,21 @@
...
+    private List<DataSet> sandboxDatasets;
+
+    public SandboxDataSets() {
```
malaffoon: nit: don't need to provide this

I'd rather a constructor that takes the collection of datasets and then make it immutable (no setter)

multi-tenant/src/test/java/org/opentestsystem/rdw/multitenant/SandboxDataSetsTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+// TODO RP-394 Move some integration test (IT) to End to End tests (ETE) and configure to run in CI
+//this test only runs against a local docker instance
+// set this property to run this test (or temporarily comment out this annotation)
+// java -Dtenant-client.docker-it-enable=true
+@Ignore
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@ActiveProfiles({"sandbox_datasets"})
+@EnableConfigurationProperties
```
malaffoon: none of lines 17-25 should be necessary for a simple POJO unit test

multi-tenant/src/test/resources/application-sandbox_datasets.yml
```diff
@@ -0,0 +1,10 @@
+logging:
+  level:
+    org.opentestsystem.rdw.admin: debug
+
```
malaffoon: Ignore my comments (that i deleted).

I thought that we were going to get the list of data sets directly from S3. Having to both drop the files and put stuff in a config file is ... onerous. Is there no way we can avoid this part?

ftdholbrook: This was was easier to code :-) we can do a directory against S3, but the upload process isn't atomic.  you could have a dataset show up in the drop down  that wasn't fully uploaded yet.  It also means every time someone goes to the sandbox login they would be hitting a slightly slower S3 call.  And one more thing that would have to be tweaked to run both locally and against S3.

malaffoon:
One question: why is this in RDW_Common, is there something in RDW_Ingest that uses it?

ftdholbrook:
> One question: why is this in RDW_Common, is there something in RDW_Ingest that uses it?

Vicky had it in admin, I had her move it.  Because it isn't tenant specific I thought it was more appropriate here.

malaffoon:
Could we put it in the common module in RDW_Reporting?

ftdholbrook:
> Could we put it in the common module in RDW_Reporting?

sure

vgonzalez9:
Closing this PR and moved these changes to reporting common.

PR #144, opened 2019-07-01
Feature/tenant configs hashcode equals

ftdholbrook:
mostly equals/hashCode/toString to tenant configurable property beans so that Set operations will work on them property (value based)

Some JsonIgnore annotations to fix serialization problems.

### DIFF
--- a/archive/src/main/java/org/opentestsystem/rdw/archive/ArchivePropertiesRoot.java
+++ b/archive/src/main/java/org/opentestsystem/rdw/archive/ArchivePropertiesRoot.java
@@ -1,5 +1,7 @@
 package org.opentestsystem.rdw.archive;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.google.common.base.Objects;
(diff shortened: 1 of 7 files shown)


archive/src/main/java/org/opentestsystem/rdw/archive/ArchivePropertiesRoot.java
```diff
@@ -89,11 +93,31 @@ public void setTenants(Map<String, ArchivePropertiesTenant> tenants) {
         this.tenants = tenants;
     }
 
+    @JsonIgnore
```
malaffoon: could we use mix-ins for this instead of annotating the class directly?

ftdholbrook: maybe?  This is the most direct route because it is used in multiple services.  We would have to modify each of those services..

or we could rename the method to status instead of getStatus so it isn't picked up during serialization.

PR #147, opened 2019-07-16
Feature/configprops events messaging

ftdholbrook:
RP-489 Alternative to making direct calls to each service to retrieve configprops


multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/configprops/ConfigpropsUpdatePublisher.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+
+    /**
+     * Publish configprops on a schedule.  Mostly a workaround for {@link ApplicationReadyEvent} not occurring
+     * late enough for cloud stream to be fully operational.  Provides a bootstrap for the central configuration
+     * that is not based configuration change events.  It will continue to publish periodically after the initial
+     * delay.
+     */
+    @Scheduled(fixedDelay = 180000L, initialDelay = 30000L)
```
esotrope: would this ever need to be tunable?

ftdholbrook: yeah that's a good idea, I think I can put EL in there

ftdholbrook: changed to use initialDelayString / fixedDelayString with EL expressions

PR #148, opened 2019-07-19
RP-601: clear caches on environment change event

gregcharles:
Services will need to import this through application configuration and provide it with a CacheManager. PR in RDW_Reporting to follow.


ftdholbrook:
This looks great Greg, my only feedback is on the name ConfigurationChangeListener.  That seems a little too generic because it listens for that change, and then evicts cache, I think putting something in the name about that would make sense.

PR #149, opened 2019-07-23
upgrade jackson to 2.9

malaffoon:
Also remove any shared version numbers in child modules

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,6 +52,14 @@ subprojects {
       dependency 'org.mockito:mockito-core:2.23.4'
       dependency 'org.mockito:mockito-inline:2.23.4'
 
+      // jackson
+      dependency 'com.fasterxml.jackson.core:jackson-annotations:2.9.9'
+      dependency 'com.fasterxml.jackson.core:jackson-core:2.9.9'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -52,6 +52,14 @@ subprojects {
...
       dependency 'org.mockito:mockito-inline:2.23.4'
 
+      // jackson
+      dependency 'com.fasterxml.jackson.core:jackson-annotations:2.9.9'
+      dependency 'com.fasterxml.jackson.core:jackson-core:2.9.9'
+      dependency 'com.fasterxml.jackson.core:jackson-databind:2.9.9.1'
+      dependency 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.9.9'
+      dependency 'com.fasterxml.jackson.datatype:jackson-datatype-jsr310:2.9.9'
+      dependency 'com.fasterxml.jackson.datatype:jackson-datatype-guava:2.9.9'
+
```
malaffoon: i built and ran all tests in reporting and ingest and there appear to be no bad side-effects

ftdholbrook: This is good, I was blocked from using Jackson for writing out in the git persistence service because of a null handling bug in 2.8.x, that means I can remove the funky Snake YAML stuff.

PR #152, opened 2019-07-25
restore null handling defaults

ftdholbrook:


vgonzalez9:
Will these still return all of the fields needed for a tenant and sandbox create?

ftdholbrook:
> Will these still return all of the fields needed for a tenant and sandbox create?

no, the UI is handling that now.

PR #153, opened 2019-08-03
remove any hint of query timeout

malaffoon:
setQueryTimeout is not well-behaved, especially when destroying statements/connections. So let's get rid of it. Also, let's override the validationInterval which defaults to 3s.

Both those things are unnecessary because our application is fairly well behaved.


multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/datasource/DataSourceElements.java
```diff
@@ -16,8 +16,6 @@
 
     DataSourceUrlParts getUrlParts();
 
-    String getJdbcInterceptors();
```
malaffoon: technically, i could've left this in but the only interceptor we're using is the QueryTimeout one

PR #157, opened 2019-08-08
SendReconciliationReportProperties support

ftdholbrook:
RP-473 multi-tenant task SendReconciliationReportProperties support
- moved ArchiveProperties to multi-tenant package to avoid circular dependency


multi-tenant/src/test/resources/application-tenant_ca.yml
```diff
@@ -44,3 +44,14 @@ task:
           oauth2:
             username: ca.user@example.com
             password: caPwd!
+  send-reconciliation-report:
+    tenants:
+      CA:
+        log: false
+        archives:
+          - uri-root: file:///tmp/
```
esotrope: woah what is this magic yaml notation?

ftdholbrook:
move the multi-tenant archive stuff to the multi-tenant package because of a circular dependency problem.  I left the package name the same, and the archive package depends on the multi-tenant package... so it should have no real impact.

multi-tenant/src/test/resources/application-tenant_ca.yml
ftdholbrook: list of objects?

ftdholbrook: curious to see how Jackson is going to handle it on writes...

esotrope: ah i see. the first one not having the other props threw me

PR #158, opened 2019-08-09
pull up tenant configurable task elements to top level

ftdholbrook:
from a code perspective not much of a change
       `@ConfigurationProperties("task.update-organizations.import-service-client")`
becomes
        `@ConfigurationProperties("task-update-organizations-import-service-client")`

some rework of the embedded application yml will be required

### DIFF
--- a/multi-tenant/src/test/resources/application-tenant_ca.yml
+++ b/multi-tenant/src/test/resources/application-tenant_ca.yml
@@ -48,6 +48,7 @@ taskSendReconciliationReport:
   tenants:
     CA:
       log: false
+      query: status=PROCESSED&after=-PT24H
       archives:
         - uri-root: file:///tmp/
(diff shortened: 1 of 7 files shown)


multi-tenant/src/test/resources/application-tenant_ca.yml
```diff
@@ -30,28 +30,28 @@ validation:
...
+  tenants:
+    CA:
+      oauth2:
+        username: ca.user@example.com
+        password: caPwd!
+
+taskSendReconciliationReport:
+  tenants:
+    CA:
+      log: false
```
malaffoon: add a test for `query`?

ftdholbrook: added query SendReconciliationReportPropertiesTenant

PR #159, opened 2019-08-15
fix for RP-709

ftdholbrook:
- Jackson mixin for ResourceOwnerPasswordResourceDetails


ftdholbrook:
I placed the mixin with the task based properties even though it is only used in the reporting admin service.  It felt like it belonged here.

ftdholbrook:
Will require a bump to common version before merging.

PR #160, opened 2019-08-16
adjust to deal with latin subject

malaffoon:


model/src/main/resources/RDW_Subject.xsd
```diff
@@ -255,13 +255,15 @@
     <xs:complexType>
       <xs:sequence>
         <!-- The performance-agnostic overall achievement message -->
```
malaffoon: The gist of the change is to make some of these optional because they aren't needed by subjects that have certain characteristics.

model/src/test/java/org/opentestsystem/rdw/common/model/subject/XmlSubjectSerializationServiceTest.java
```diff
@@ -159,6 +159,23 @@ public void itShouldLoadELPAC() {
         }
     }
 
+    @Test
+    public void itShouldLoadLatin() {
+        try (final InputStream is = this.getClass().getResourceAsStream("/Latin.xml")) {
+            final Subject subject = service.parseSubject(is);
```
malaffoon: Mostly the test is just parsing the file: it was failing before these changes

PR #161, opened 2019-08-21
RP-719 task send reconciliation report

ftdholbrook:
- add resolver to provide default for query

### DIFF
--- a/multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/task/SendReconciliationReportPropertiesTenant.java
+++ b/multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/task/SendReconciliationReportPropertiesTenant.java
@@ -6,11 +6,12 @@ import org.opentestsystem.rdw.archive.ArchivePropertiesTenant;
 import java.util.ArrayList;
 import java.util.List;
 
-public class SendReconciliationReportPropertiesTenant {
+public class SendReconciliationReportPropertiesTenant implements SendReconciliationReportProperties {
     private List<ArchivePropertiesTenant> archives = new ArrayList<ArchivePropertiesTenant>();
-    private boolean log;
+    private Boolean log;
     private String query;
 
+    @Override
     public List<ArchivePropertiesTenant> getArchives() {
         return archives;
     }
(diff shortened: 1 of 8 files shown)


ftdholbrook:
This is the first PR of 3, also Ingest and Reporting changes incoming.

multi-tenant/src/main/java/org/opentestsystem/rdw/multitenant/task/SendReconciliationReportPropertiesTenant.java
```diff
@@ -6,11 +6,12 @@
 import java.util.ArrayList;
 import java.util.List;
 
-public class SendReconciliationReportPropertiesTenant {
+public class SendReconciliationReportPropertiesTenant implements SendReconciliationReportProperties {
     private List<ArchivePropertiesTenant> archives = new ArrayList<ArchivePropertiesTenant>();
-    private boolean log;
+    private Boolean log;
```
ftdholbrook: required nullalble for override

multi-tenant/src/test/resources/application.yml
```diff
@@ -72,4 +72,8 @@ taskUpdateOrganizationsImportServiceClient:
   oauth2:
     access-token-uri: https://sso-deployment.sbtds.org:443/auth/oauth2/access_token?realm=/sbac
     client-id: sbacdw
-    client-secret:
\ No newline at end of file
+    client-secret:
+
+taskSendReconciliationReport:
+  query: DEFAULT QUERY
+  log: false
```
ftdholbrook: default to false

malaffoon: Does it default to false without this? Or do i have to document it as required to set it false?

ftdholbrook: it will default to null without it, in ingest I set it in the application.yml in the jar file, that should handle it.  No need for external config.

malaffoon: i don't see it in the application.yml file; probably user-error but ... ?

ftdholbrook: https://github.com/SmarterApp/RDW_Ingest/pull/647/files#diff-4f1e9c094a7095bf53bb36a68284d945R75

PR #163, opened 2019-09-14
pipeline error memory test

malaffoon:


script/src/main/java/org/opentestsystem/rdw/script/PipelineScript.java
```diff
@@ -1,16 +1,13 @@
...
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.Map;
 
 /**
  * Base class for pipeline scripts. Implementation will be loaded from Groovy script file.
  */
 public abstract class PipelineScript extends Script {
-    protected static final Logger logger = LoggerFactory.getLogger(PipelineScript.class);
```
malaffoon: unrelated warning fix

script/src/test/java/org/opentestsystem/rdw/script/DSLScriptBaseIT.java
```diff
@@ -66,4 +67,33 @@ public void itShouldNotThrowExceptionForExistingProperty() throws Exception {
         assertThat(results.asString()).isEqualTo(Boolean.TRUE.toString());
     }
 
+    @Test
```
malaffoon: So this demonstrates the problem. Still thinking about the best way to fix it.

It feels like it should be in Pipeline since that class knows a little bit about system scripts. But not sure if that is right.

Should it be fixed with some sort of a `reset` or `init` hook in the script? How does the java framework invoke that? I guess a quick fix would be to include a line to clear the errorCollector at the beginning of the post-process.groovy script (kind of a bookend for the checkValid call). @gregcharles, thoughts?

malaffoon: FTR, i can make the test work if i change the script to be this:
```
parserHelper.elementErrorCollector.reset()
if (input == 'error') addError 'input', input, 'snafu'
checkValid
```

script/src/main/resources/scripts/DSLScriptBase.groovy
```diff
@@ -83,6 +87,24 @@ abstract class DSLScriptBase extends PipelineScript {
...
+     * This method is defined the way it is so it can be used more simply in the script:<pre>
+     * resetValid
+     * # do some stuff
+     * checkValid
+     * </pre>
+     * </p>
+     *
+     * @see DSLScriptBase#getCheckValid
+     */
+    def getResetValid() {
```
malaffoon: This will work but it is prone to error since it is a "cooperative" fix.

@gregcharles, is declaring the method this way the only way to allow them to use `resetValid` in the script? I really don't like having a "getter" with side effects but i also don't want them to have to put `resetValid()`.

malaffoon:
Fixed elsehow

PR #164, opened 2019-09-17
RP-809: restore instantiation of script object for each pipeline run

gregcharles:
This implementation is a bit unusual as it takes a script object and uses it as a prototype to create new script objects. Eventually, I'd  prefer to restore  the implementation that created  the Pipeline with the script classes and configuration, rather than script objects.  However, this will take more intensive refactoring and testing.

### DIFF
--- a/script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
+++ b/script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
@@ -157,7 +157,9 @@ public class Pipeline {
     }
 
     /**
-     * Get an instance of the script to be run against the given input
+     * Get an instance of the script to be run against the given input.
+     * This constructs a copy of the script with the original bound properties and source
+     * to guarantee we have a fresh instance with no memory of previous processing.
      *
      * @param prototype prototype instance of the Pipeline script
      * @param input input for the script to work on. The type is script dependent.
--- a/script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
+++ b/script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
@@ -1,6 +1,6 @@
 package org.opentestsystem.rdw.script;
 
-import org.opentestsystem.rdw.script.security.AbstractToggleableSecurityManager;
+import groovy.lang.MissingPropertyException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -12,6 +12,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import org.opentestsystem.rdw.script.security.AbstractToggleableSecurityManager;
 
 import static com.google.common.collect.ImmutableList.copyOf;
 
@@ -29,6 +30,7 @@ public class Pipeline {
     private final List<PipelineScript> scripts;
     private final AbstractToggleableSecurityManager sandboxSecurityManager;
 
+    // TODO: restore the implementation that passes a list of PipelineDefinitions instead of PipelineScripts.
     public Pipeline(
             @NotNull final PipelineDefinition source,
             @NotNull final List<PipelineScript> scripts,
@@ -61,8 +63,10 @@ public class Pipeline {
 
         Object lastValidTransformation = input;
 
-        for (final PipelineScript script : scripts) {
+        for (final PipelineScript prototype : scripts) {
+            PipelineScript script = null;
             try {
+                script = getInstance(prototype, lastValidTransformation);
                 script.bindInput(lastValidTransformation);
 
                 final Object results = runScript(script);
@@ -152,4 +156,27 @@ public class Pipeline {
         return null;
     }
 
+    /**
+     * Get an instance of the script to be run against the given input
+     *
+     * @param prototype prototype instance of the Pipeline script
+     * @param input input for the script to work on. The type is script dependent.
+     * @return an instance of a PipelineScript
+     * @throws Exception if there is an problem constructing the script
+     */
+    private PipelineScript getInstance(final PipelineScript prototype, final Object input) throws Exception {
+        // TODO: this method should be updated to work with PipelineDefinition, not PipelineScript
+        final PipelineScript script = prototype.getClass().getConstructor().newInstance()
+                .bindProperties(prototype.getBoundProperties())
+                .source(prototype.getSource())
+                .bindInput(input);
+
+        try {
+            script.setProperty("propertyResolver", prototype.getProperty("propertyResolver"));
+        } catch (MissingPropertyException e) {
+            logger.info("No property resolver set");
+        }
+
+        return script;
+    }
 }
--- a/script/src/main/java/org/opentestsystem/rdw/script/PipelineScript.java
+++ b/script/src/main/java/org/opentestsystem/rdw/script/PipelineScript.java
@@ -1,12 +1,9 @@
 package org.opentestsystem.rdw.script;
 
-import com.google.common.collect.ImmutableMap;
 import groovy.lang.Script;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import javax.print.attribute.HashDocAttributeSet;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -17,7 +14,7 @@ public abstract class PipelineScript extends Script {
     protected static final Logger logger = LoggerFactory.getLogger(PipelineScript.class);
 
     private PipelineScriptDefinition source;
-    private Map<String,Object> boundProperties = Collections.EMPTY_MAP;
+    private Map<String,Object> boundProperties = new HashMap<>();
 
     public PipelineScriptDefinition getSource() {
         return source;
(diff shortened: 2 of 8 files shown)


script/src/main/java/org/opentestsystem/rdw/script/PipelineScript.java
```diff
@@ -1,9 +1,13 @@
 package org.opentestsystem.rdw.script;
 
+import com.google.common.collect.ImmutableMap;
 import groovy.lang.Script;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.print.attribute.HashDocAttributeSet;
```
malaffoon: ?!

script/src/test/java/org/opentestsystem/rdw/script/TextPipelineTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+        when(repository.findByCodeAndVersion(PIPELINE_NAME, PIPELINE_VERSION))
+                .thenReturn(pipelineWithUserScript(scriptCode));
+
+        Pipeline pipeline = pipelineFactory.getPipeline(PIPELINE_NAME, PIPELINE_VERSION);
+
+        try {
+            pipeline.run("good input");
+        } catch (ScriptRuntimeException sre) {
+            throw (Exception)sre.getCause();
```
malaffoon: This seems a little ... indirect ... for a test. Are you testing that the sre cause is expected to be an ImportException? Perhaps use assertThatThrownBy or something similar?

script/src/test/java/org/opentestsystem/rdw/script/TextPipelineTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+        Pipeline pipeline = pipelineFactory.getPipeline(PIPELINE_NAME, PIPELINE_VERSION);
+
+        try {
+            pipeline.run("bad input");
+            fail("Expected a script exception for bad input");
+        } catch (ScriptRuntimeException sre) {
+        }
+
+        pipeline.run("good input");
```
malaffoon: This should really be asserting something ... perhaps just `assertThat(pipeline.run("good input").asString()).isEqualTo("good input");`

script/src/test/java/org/opentestsystem/rdw/script/TextPipelineTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+        when(repository.findByCodeAndVersion(PIPELINE_NAME, PIPELINE_VERSION))
+                .thenReturn(pipelineWithUserScript(scriptCode));
+
+        Pipeline pipeline = pipelineFactory.getPipeline(PIPELINE_NAME, PIPELINE_VERSION);
+
+        try {
+            pipeline.run("bad input");
+            fail("Expected a script exception for bad input");
+        } catch (ScriptRuntimeException sre) {
```
malaffoon: Again, you could use assertThatThrownBy or something similar ... it makes the test read better by being explicit about what is expected.

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
```diff
@@ -152,4 +156,27 @@ private ScriptRuntimeException extractCause(final Throwable t) {
...
 
+    /**
+     * Get an instance of the script to be run against the given input
+     *
+     * @param prototype prototype instance of the Pipeline script
+     * @param input input for the script to work on. The type is script dependent.
+     * @return an instance of a PipelineScript
+     * @throws Exception if there is an problem constructing the script
+     */
+    private PipelineScript getInstance(final PipelineScript prototype, final Object input) throws Exception {
```
malaffoon: It feels wrong to me that this isn't in the PipelineFactory somehow. I guess one could argue that this is the moral equivalent of a copy-constructor?

script/src/main/java/org/opentestsystem/rdw/script/PipelineScript.java
```diff
@@ -31,15 +36,35 @@ public PipelineScript source(final PipelineScriptDefinition source) {
...
+            addBoundProperties(properties);
             for (Map.Entry<String,Object> entry : properties.entrySet()) {
                 this.setProperty(entry.getKey(), entry.getValue());
             }
         }
-
         return this;
     }
 
+    private void addBoundProperties(final Map<String, Object> properties) {
```
malaffoon: i sort of see the desire to have an immutable map but this is complex for something that is used only internally. That is, this entire method could be replaced with a single line up around line 40.

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
```diff
@@ -152,4 +156,27 @@ private ScriptRuntimeException extractCause(final Throwable t) {
         return null;
     }
 
+    /**
+     * Get an instance of the script to be run against the given input
+     *
```
malaffoon: Some additional javadoc would be nice, perhaps
```
 * This constructs a copy of the script with the original bound properties and source
 * to guarantee we have a fresh instance with no memory of previous processing.
```

malaffoon:
Hmm, i'm waffling on this one. It feels like we may want to do the short term "fix" to checkValid, and then leave the (techdebt) JIRA around to refactor it properly as you describe. Meh.

script/src/main/java/org/opentestsystem/rdw/script/PipelineScript.java
gregcharles: Oops, imports now optimized. (It was a suggestion for Hash... that was HashMap, when finger moved toward the Enter key, but HasDocAttributeSet when it landed. IntelliJ is a merry prankster!)

script/src/test/java/org/opentestsystem/rdw/script/TextPipelineTest.java
gregcharles: done

script/src/test/java/org/opentestsystem/rdw/script/TextPipelineTest.java
gregcharles: done

script/src/main/java/org/opentestsystem/rdw/script/PipelineScript.java
gregcharles: done

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
gregcharles: Done

script/src/test/java/org/opentestsystem/rdw/script/TextPipelineTest.java
gregcharles: done

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
gregcharles: I'll put in a technical debt ticket to track this. My preference is for the construction to only happen in Pipeline, but that will require more extensive changes.

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
```diff
@@ -152,4 +156,27 @@ private ScriptRuntimeException extractCause(final Throwable t) {
...
+     * @param input input for the script to work on. The type is script dependent.
+     * @return an instance of a PipelineScript
+     * @throws Exception if there is an problem constructing the script
+     */
+    private PipelineScript getInstance(final PipelineScript prototype, final Object input) throws Exception {
+        // TODO: this method should be updated to work with PipelineDefinition, not PipelineScript
+        final PipelineScript script = prototype.getClass().getConstructor().newInstance()
+                .bindProperties(prototype.getBoundProperties())
+                .source(prototype.getSource())
+                .bindInput(input);
```
malaffoon: And you have absolutely no performance concerns with this change? I know exam processing has I/O that is the long pole but this is a lot of introspection and object construction for every exam being processed.

script/src/main/java/org/opentestsystem/rdw/script/PipelineScript.java
malaffoon: i'm confused, it's still there?

script/src/main/java/org/opentestsystem/rdw/script/Pipeline.java
gregcharles: I'm worried more with the reflection than the construction. Nothing against Spring, but I'd be surprised if there weren't several hundred objects created for each run of an exam. I suppose I could cache the constructors the first time I reflected for them. Still, I just tested 100,000 runs of reflect, construct, and bind input and that ran in 1.1 seconds. That didn't include binding other properties, but it did include printing out the instance references. (It ran so fast without the print that I was afraid it was just optimizing away the whole loop.)

gregcharles: `    public void performance() throws Exception {
        final String scriptCode = "enable('csv')\n" +
                "transformRows {row ->\n" +
                "   def value = row.bogus\n" +
                "}\n" +
                "outputCsv";

        final Class<? extends PipelineScript> scriptClass = compiler.compile(scriptCode);
        long start = System.currentTimeMillis();

        for (int i = 0; i < 100000; i++) {
            System.out.println(scriptClass.getConstructor().newInstance().bindInput(csv));
        }

        System.out.println(System.currentTimeMillis() - start);
    }
`

malaffoon: Ok, thanks for taking the time to play with that. I'll avoid premature optimization ;)

PR #165, opened 2019-10-08
RP-826: enhancement to allow internal commas in tenancy chain

gregcharles:


utils/src/main/java/org/opentestsystem/rdw/utils/TenancyChain.java
```diff
@@ -60,14 +60,40 @@ public static TenancyChain fromString(final String value) {
...
+        if (grants.length <= 1) {
+            return grants;
+        }
+
+        // Restores pipe characters lost because of the split.
+        grants[0] = grants[0] + '|';
+        for (int i = 1; i < grants.length - 1 ; i++) {
+            grants[i] = '|' + grants[i] + '|';
+        }
+        grants[grants.length-1] = '|' + grants[grants.length-1];
```
malaffoon: Having to do this mess kind of sucks. But the only alternative i can think of is to not use the default `split` function and do it explicitly with a pattern matcher. Which probably isn't any more maintainable:
```
    public String[] splitGrants(final String value) {
        final List<String> splits = newArrayList();
        final Matcher matcher = Pattern.compile("\\|,\\|").matcher(value);
        int index = 0;
        while (matcher.find()) {
            splits.add(value.substring(index, matcher.start() + 1));
            index = matcher.end() - 1;
        }
        if (index == 0) {
            return new String[]{ value };
        } else {
            splits.add(value.substring(index));
            return splits.toArray(new String[0]);
        }
    }
```

malaffoon: Could improve that a teeny bit more by making the compiled pattern a class variable.

gregcharles: I'm not sure about maintainability, but the pattern matcher way seems way more cryptic to me. I agree that the first-middle-last change is a bit messy though, so if you feel strongly about it, I'll amend. I going to put it in as-is for now though.

PR #169, opened 2020-06-04
add new colors; add alt-score level const; tweak gradle

malaffoon:
I tried loading a subject with the new colors and it failed because of schema validation. This adds the colors to the xsd. And a couple other tweaks.


build.gradle
```diff
@@ -91,32 +91,20 @@ subprojects {
         finalizedBy jacocoTestReport
     }
 
-    task sourcesJar(type: Jar, dependsOn: classes) {
```
malaffoon, line 94: gradle 6 simplifies creating source and doc jars

model/src/main/java/org/opentestsystem/rdw/common/model/AssessmentBoundaries.java
```diff
@@ -13,8 +13,12 @@
...
-     * The maximum number of scorable claims that are allowed for
-     * any subject.
+     * The maximum number of scorable claims that are allowed for any subject.
      */
     public static final int MaxScorableClaims = 6;
+
+    /**
+     * The maximum number of alt scores that are allowed for any subject.
+     */
+    public static final int MaxAltScores = 6;
```
malaffoon, line 23: need this for alt-score aggregate reporting

model/src/main/java/org/opentestsystem/rdw/common/model/subject/XmlSubjectSerializationService.java
```diff
@@ -34,10 +35,14 @@
...
-        setDataOrder(subject);
-        return subject;
+        try {
+            final Subject subject = getMapper().readValue(getStreamReader(inputStream), Subject.class);
+            setTargetReport(subject);
+            setDataOrder(subject);
+            return subject;
+        } catch (final JsonProcessingException e) {
+            throw new IllegalArgumentException(e);
+        }
```
malaffoon, line 45: This will, very slightly, change behavior. For schema violations, we are converting the exception to an IAE because the system knows how to catch those and be friendly about it. Previously, schema violations were treated as low-level IOExceptions ... less friendly.

model/src/main/resources/RDW_Subject.xsd
```diff
@@ -495,16 +495,36 @@
   <!-- The colors available within the RDW system -->
   <xs:simpleType name="Color">
     <xs:restriction base="xs:string">
-      <xs:enumeration value="blue"/>
```
gregcharles, line 498: Yikes, I forgot to merge these! Thanks for picking them up.

PR #170, opened 2020-06-05
replace spring-boot-starter-data-jpa with -jdbc, not really sure why …

malaffoon:
…i made it "data-jpa" to begin with


group/build.gradle
```diff
@@ -7,7 +7,7 @@ dependencies {
     implementation 'org.apache.commons:commons-csv'
     implementation 'org.apache.commons:commons-lang3'
     implementation 'org.apache.commons:commons-text'
-    implementation 'org.springframework.boot:spring-boot-starter-data-jpa'
+    implementation 'org.springframework.boot:spring-boot-starter-jdbc'
```
malaffoon, line 10: spring-boot-starter-data-jpa was dragging in all sorts of hibernate cruft which we definitely don't want.

gregcharles: Gah, that's exactly what I found on SO, but I searched for spring-boot-starter-data-jpa on RDW_Reporting and since it wasn't there, figured there must have been something else wrong. Well, I feel better that I got close at least.

PR #172, opened 2020-06-10
add subject traits

malaffoon:
This enhances the (common) subject model to add traits.


model/src/main/java/org/opentestsystem/rdw/common/model/subject/Trait.java
```diff
@@ -0,0 +1,70 @@
+package org.opentestsystem.rdw.common.model.subject;
+
+/**
+ * This class represents subject trait codes.
+ * A trait code is used to identify additional scores for an exam.
+ * The other trait attributes (purpose, category) are used for grouping score
+ * values in reports. The other traits don't have to match the code text but
+ * usually do.
+ *
```
gregcharles, line 9: What do you mean by "other traits" here? Other than which?

model/src/main/java/org/opentestsystem/rdw/common/model/subject/Trait.java
```diff
@@ -0,0 +1,70 @@
...
+ * values in reports. The other traits don't have to match the code text but
+ * usually do.
+ * 
+ * For example, ELA has traits for five purposes, each with three categories.
+ * SOCK_ARGU_ORG is the code that identifies the score for organization (ORG)
+ * of the answer to the question with purpose argument (ARGU). Typically, a 
+ * test result has one or more trait scores for a single purpose. In this 
+ * example, a test result may have three ARGU scores, SOCK_ARGU_ORG, 
+ * SOCK_ARGU_CON, SOCK_ARGU_EVI.
+ *
```
gregcharles, line 16: Now actually I'm wondering where SOCK comes from, when just using ELA would be more natural.  Are explicit codes even useful? Could they just be constructed from subject, category, and purpose?

model/src/main/java/org/opentestsystem/rdw/common/model/subject/Trait.java
malaffoon: Oh, bad choice of words on my part. By "other traits" i mean other attributes of the Trait object, specifically the purpose/category. I'll rephrase.

model/src/main/java/org/opentestsystem/rdw/common/model/subject/Trait.java
malaffoon: I'm really not sure the origin of SOCK but SmarterBalanced has that as a prefix for ELA-related claims. 

I have taken the path of making the code explicitly declared. Relying on a particular internal format of the code will be problematic. A lesson i learned a long time ago is to treat values as opaque. So i'm making them declare the code, they can construct it however they want.

